PR #5, opened 2026-03-14
browseDirectory protocol addition

roblourens:
## Summary
This brings over the protocol-facing changes from microsoft/vscode commit `ce5e5036406abb45311cec952552f0fbaeda550a` into the agent-host-protocol website.

## Changes
- add `defaultDirectory?` to the `initialize` result shape
- add `browseDirectory(uri)` command types and generated command schema
- regenerate the commands and messages reference pages
- update manual lifecycle, overview, and getting started docs to mention the new handshake field and request

## Validation
- `npm run generate`
- `npm run typecheck`
- `npm run docs:build`

(Written by Copilot)


types/commands.ts
```diff
@@ -248,6 +250,43 @@ export interface IFetchContentResult {
...
+  /** Base name of the entry */
+  name: string;
+  /** Whether the entry is a file or directory */
+  type: 'file' | 'directory';
+}
+
+/**
+ * Result of the `browseDirectory` command.
+ */
+export interface IBrowseDirectoryResult {
```
connor4312, line 289: We should spec a behavior if the directory doesn't exist / doesn't have permissions

connor4312: I saw this commit pop up but I think you or the agent might have lost it 😛 

`AhpErrorCodes` in errors.ts would be the place to put any error code

roblourens: Would you have specific error codes for these failures cases?

docs/reference/commands.md
```diff
@@ -229,7 +230,44 @@ use `utf-8` encoding.
...
+## `browseDirectory` <a href="https://github.com/microsoft/agent-host-protocol/blob/main/types/commands.ts#L271" title="View source" style="float:right;font-size:0.75em;opacity:0.5;text-decoration:none">📄</a>
+
+Lists directory entries at a file URI on the server's filesystem.
+
+This is intended for remote folder pickers and similar UI that needs to let
+users navigate the server's local filesystem.
+
+The server MUST return success only if the target exists and is a directory.
+If the target does not exist, is not a directory, or cannot be accessed, the
+server MUST return a JSON-RPC error.
```
connor4312, line 242: We should define a specific error code for this

https://github.com/microsoft/agent-host-protocol/blob/07a52068358f12f3265c6102b4078a5060859538/types/errors.ts#L38

connor4312 (approved):

rzhao271 (approved):

PR #32, opened 2026-03-30
Add settings support to the protocol

roblourens:
Introduce a settings system with server-advertised schema and typed client-pushed values.

## Design

Two complementary pieces on root state:

1. **`settingsSchema?: ISettingsSchema`** — server-advertised JSON Schema describing which settings the host supports, their types, defaults, and descriptions. Updated via `root/settingsSchemaChanged`. Clients use this to render a generic settings editor.

2. **`settings?: IAgentHostSettings`** — client-pushed setting values. Updated via `root/settingsChanged` (full-replace semantics). Well-known keys are explicitly typed for compile-time safety; an index signature allows dynamic settings the server may advertise at runtime.

Session-level overrides are supported: `ISessionState.settings` can override root settings for settings whose schema has `scope: 'session'`.

## New Types

- **`SettingScope`** enum (`'root'` | `'session'`) — declares where a setting can be configured
- **`ISettingPropertySchema`** — recursive JSON Schema subset for setting value types (no `scope`)
- **`ISettingRegistration extends ISettingPropertySchema`** — adds `scope` at the top level only
- **`ISettingsSchema`** — `type: 'object'` with `properties?: Record<string, ISettingRegistration>`
- **`IAgentHostSettings`** — typed well-known keys + `[key: string]: unknown` for dynamic ones

## New Actions

| Action | Direction | Description |
|--------|-----------|-------------|
| `root/settingsSchemaChanged` | Server → Client | Advertise supported settings schema |
| `root/settingsChanged` | Client → Server | Push root-level setting values |
| `session/settingsChanged` | Client → Server | Push per-session setting overrides |

## First Well-Known Setting

`tools.edits.autoApprove` (`Record<string, boolean>`) — glob-pattern map controlling which file-edit tool calls are auto-approved. Equivalent to VS Code's `chat.tools.edits.autoApprove`. Declared with `scope: 'session'` so it can be overridden per-session.

## Other

- Fix in `generate-json-schema.ts` to strip quotes from string-literal property names (e.g. `'tools.edits.autoApprove'` was being emitted as `"'tools.edits.autoApprove'"` in generated schemas)

(Written by Copilot)

### DIFF
--- a/types/actions.ts
+++ b/types/actions.ts
@@ -10,6 +10,8 @@ import type {
   URI,
   StringOrMarkdown,
   IAgentInfo,
+  IAgentHostSettings,
+  ISettingsSchema,
   IErrorInfo,
   IUserMessage,
   IResponsePart,
@@ -32,6 +34,8 @@ import { ToolCallConfirmationReason, ToolCallCancellationReason, PendingMessageK
 export const enum ActionType {
   RootAgentsChanged = 'root/agentsChanged',
   RootActiveSessionsChanged = 'root/activeSessionsChanged',
+  RootSettingsSchemaChanged = 'root/settingsSchemaChanged',
+  RootSettingsChanged = 'root/settingsChanged',
   SessionReady = 'session/ready',
   SessionCreationFailed = 'session/creationFailed',
   SessionTurnStarted = 'session/turnStarted',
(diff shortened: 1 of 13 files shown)


types/actions.ts
```diff
@@ -130,6 +135,39 @@ export interface IRootActiveSessionsChangedAction {
...
+ * {@link IRootState.settingsSchema | settingsSchema}.
+ *
+ * @category Root Actions
+ * @version 1
+ * @clientDispatchable
+ */
+export interface IRootSettingsChangedAction {
+  type: ActionType.RootSettingsChanged;
+  /** The complete settings values object */
+  settings: IAgentHostSettings;
```
connor4312, line 152: This could be `Partial<IAgentHostSettings>`, I don't think there's need to send the whole object down again. Over time the object will likely get kinda big.

PR #35, opened 2026-04-01
Changes for showing edits better and more resource/fs operations

connor4312:
A set of changes that allows representing edits completely, and also expands and makes fs/resource operations more consistent.


roblourens (approved):
Do you need something like a stat? For checking existence and getting metadata without reading the whole thing? 

Maybe partial reading? Write with append?

connor4312:
All reasonable things. In practice though I'm not sure how comprehensive our FS operations will need to be with the scope of AHP. I opted to keep it simple thus far until we start seeing places where we'd want those 🤷

PR #77, opened 2026-04-23
support remote host plugin configuration metadata

joshspicer:
## Summary
- allow root config actions to flow through the client-dispatch typing/generation path (`ClientRootAction` / `ServerRootAction` unions)
- document how host-owned customizations live in root config for remote agent hosts
- use `clientId` on `SessionCustomization` to distinguish host-provided vs client-provided plugins

## Two categories of customizations

1. **Host customizations** — Plugins configured on the remote host's filesystem. Published as `SessionCustomization` entries **without** `clientId`.

2. **Client customizations** — Plugins the active client contributes via `session/activeClientChanged`. Published as `SessionCustomization` entries **with** `clientId` set to the contributing client.

## Why
Remote/browser clients may not have local customizations to sync into a connected agent host. This change gives the protocol enough structure to represent host-owned plugin configuration, which the paired VS Code change uses to configure plugins directly on the remote AHP host.

## Paired implementation
Paired draft PR: microsoft/vscode#312225

## Notes for reviewers
- this is the protocol/spec half of the feature
- `SessionCustomizationSource` enum removed per review feedback — `clientId` field is sufficient
- `CustomizationScopeKind` / `scope` removed per review feedback — workspace scoping is not needed at protocol level
- the paired VS Code implementation wires these fields into persisted root config, host side effects, and the Sessions customization UI


types/state.ts
```diff
@@ -1453,16 +1483,27 @@ export const enum CustomizationStatus {
...
+ * Origin of a customization active in a session.
  *
- * Entries without a `clientId` are server-provided; entries with a `clientId`
- * originate from that client.
+ * @category Customization Types
+ */
+export const enum SessionCustomizationSource {
+  /** Customization was configured by the connected client. */
+  Client = 'client',
+  /** Customization was configured on the agent host itself. */
```
connor4312: `SessionCustomization` has a `clientId?: string` that should be sufficient to distinguish this I think -- just check that `clientId === undefined` to know if it's host-provided

types/state.ts
```diff
@@ -1409,6 +1409,34 @@ export type ToolResultContent =
...
+ */
+export interface CustomizationScope {
+  /** Whether the customization is host-wide or workspace-specific. */
+  kind: CustomizationScopeKind;
+  /**
+   * Workspace root URI this customization applies to.
+   *
+   * Required when {@link kind} is {@link CustomizationScopeKind.Workspace}.
+   */
+  workspace?: URI;
```
connor4312: Since customization refs are currently _only_ in the scope of a session, do we need this?

Is the scenario that you might have a session whose working directory is inside a `workspace` directory for which this customization applies?

joshspicer: good point, I don't think I need this.

dmitrivMS (approved):

connor4312 (approved):

PR #88, opened 2026-04-27
Add Agent Host Protocol threat model

rwoll:
## Summary
- adds a concise THREAT_MODEL.md for Agent Host Protocol implementations
- frames the desired untrusted mode goal for remote/SSH/tunnel usage
- documents core risks around hostile peers, token forwarding, client tools, resource APIs, terminals, plugins, packages, and multi-client ownership

## Testing
- documentation-only change

### DIFF
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -2,6 +2,10 @@
 
 This document captures the core security assumptions, risks, and minimum controls for Agent Host Protocol (AHP) implementations. AHP is a bidirectional JSON-RPC protocol that lets clients connect to an agent host, subscribe to session state, dispatch actions, access resources, create terminals, authenticate to protected resources, and optionally contribute client-side tools and customizations.
 
+## Current safety status
+
+The current protocol **is not safe by itself for untrusted remote use**. AHP defines the message shapes and state flows, but it does not itself establish peer identity, authorize capabilities, constrain token forwarding, sandbox resource access, sanitize rendered content, or make client-contributed tools safe to invoke. Implementations must add those controls before treating a remote host or client as safe.
+
 ## Goal: untrusted mode
 
 The desired end state is an **untrusted mode**: AHP and client implementations should let a user connect to an agent host and still get useful work products without risking local compromise or long-lived secrets. In this mode, the client treats the host as hostile by default, disables or tightly scopes client tools, avoids silent token forwarding, and only shares explicit, bounded inputs.
--- a/THREAT_MODEL.md
+++ b/THREAT_MODEL.md
@@ -1,0 +1,86 @@
+# Threat Model
+
+This document captures the core security assumptions, risks, and minimum controls for Agent Host Protocol (AHP) implementations. AHP is a bidirectional JSON-RPC protocol that lets clients connect to an agent host, subscribe to session state, dispatch actions, access resources, create terminals, authenticate to protected resources, and optionally contribute client-side tools and customizations.
+
+## Goal: untrusted mode
+
+The desired end state is an **untrusted mode**: AHP and client implementations should let a user connect to an agent host and still get useful work products without risking local compromise or long-lived secrets. In this mode, the client treats the host as hostile by default, disables or tightly scopes client tools, avoids silent token forwarding, and only shares explicit, bounded inputs.
+
+This mode is still valuable. A user might only want the agent to produce artifacts such as a `PLAN.md`, or might give the host a short-lived, least-privilege token that can push a PR as an external, non-collaborator identity. The threat model below describes current implementation risks so the protocol and clients can work toward this goal while preserving a hard boundary around the user's local client, local files, credentials, and durable account authority.
+
+## Security model
+
+- **AHP messages are untrusted data, not authority.** A peer can spoof agent names, safety text, auth prompts, tool confirmations, URLs, terminal output, file diffs, and session state.
+- **Transport identity is outside the wire protocol.** WebSocket, SSH, tunnel, MessagePort, or custom transports must authenticate peers before `initialize`. AHP `clientId` is not an identity unless it is bound to the authenticated transport.
+- **Remote hosts are high-risk by default.** In SSH/tunnel/remote WebSocket scenarios, assume the remote agent host, agent SDK, model output, workspace, terminal environment, package installs, and all server messages may be compromised.
+- **A compromised remote host is not automatically a compromised local client.** The local client becomes exposed when it renders hostile content, forwards tokens, serves local resources, or executes client-contributed tools.
+- **Autonomous or "YOLO" agents expand the blast radius.** If an agent can run shell commands or install packages without review, treat the remote workspace, remote environment variables, package manager credentials, and tokens already sent to the host as compromised.
+
+## SSH / tunnel placement
+
+```mermaid
+flowchart LR
+    subgraph Local["Client machine"]
+        Client["AHP client"]
+        Tools["Client tools"]
+        Secret["Client token / secret"]
+        Relay["SSH / tunnel / WebSocket client"]
+    end
+
+    subgraph Link["Transport"]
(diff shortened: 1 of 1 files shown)


rwoll:
@connor4312 @roblourens - here's my first draft. for the purposes of MSRC, it's important we are all on the same page on accepted risks. so please review and comment if you see anything factually incorrect, or that we want to fix. longer term, I do think it's critical we have an "untrusted" mode. Some of these risks (e.g. forwarding a credential to a client) will likely need UX treatments as well as public docs for education.

roblourens (approved):
This makes sense to me, thanks for digging into it

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
+# Threat Model
+
+This document captures the core security assumptions, risks, and minimum controls for Agent Host Protocol (AHP) implementations. AHP is a bidirectional JSON-RPC protocol that lets clients connect to an agent host, subscribe to session state, dispatch actions, access resources, create terminals, authenticate to protected resources, and optionally contribute client-side tools and customizations.
+
+## Current safety status
+
+The current protocol **is not safe by itself for untrusted remote use**. AHP defines the message shapes and state flows, but it does not itself establish peer identity, authorize capabilities, constrain token forwarding, sandbox resource access, sanitize rendered content, or make client-contributed tools safe to invoke. Implementations must add those controls before treating a remote host or client as safe.
```
connor4312, line 7: I would actually say the opposite. The protocol is just RPC and it doesn't have an "execute locally" command. A vanilla implementation will be 'safe' by default unless a client gives it e.g. local tools it could use to escape.

Maybe it's the wording "the current protocol" -- there are no protocol changes we would make that would make it safer than it is now.

We might add things like sandboxing in the future but this is an optional implementation, it doesn't make the protocol inherently more or less safe.

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+| **Client-contributed tool abuse** | The active client contributes tools; the server starts a tool call with `toolClientId` and attacker-controlled input that causes the client to run tasks, inspect local context, or return sensitive output. | Client tools must be explicit, allowlisted, capability-scoped, and locally authorized. Treat server `confirmed: "not-needed"` as advisory, not as client approval. Confirm tools that read/write local data, run commands, open URLs, or use credentials. |
+| **Server-side workspace compromise by malicious clients** | An unauthorized client invokes `resourceWrite`, `resourceDelete`, terminal creation/input, tool confirmations, or config/customization actions against the host. | Authenticate and authorize each client independently. Bind `clientId` to the transport identity. Canonicalize resource URIs, sandbox filesystem access, gate terminal operations, and audit privileged actions. |
```
connor4312, line 73: There is the converse as well -- resource commands are bidirectional. Clients should be intentional about the URIs it allows the server to interop with, such as by restricting them to only paths/directories the clients has already announced in a ContentRef

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+| **WebSocket or tunnel exposure** | An agent host listens on a reachable interface; a browser, local malware, or network attacker connects and sends AHP commands. | Bind loopback by default, require explicit opt-in for network exposure, authenticate during WebSocket upgrade, use `wss` for remote connections, enforce Origin checks for browser-reachable endpoints, and rate-limit. |
```
connor4312, line 74: "Origin checks for browser-reachable endpoints, and rate-limit" is not sufficient to protect loopbacks. Loopbacks should require a randomized/secure token in their URIs to make them unguessable

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+| **Multi-client confusion** | One client races another to claim active-client status, approve a tool call, complete a client tool, or write terminal input. | Authorize every action against current server state and authenticated connection identity. Scope approvals, terminal claims, and tool completion to the owning client. Reject stale or replayed decisions. |
```
connor4312, line 75: >Scope approvals, terminal claims, and tool completion to the owning client

Multi-client support is the goal of AHP in the first place. I would just drop this row.

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+| **Plugin, customization, and package supply-chain execution** | A host loads a remote customization or a YOLO agent runs `npm install` / `pip install`; install scripts or plugin code read secrets and exfiltrate them. | Treat plugin loading and package installation as code execution. Require provenance, signatures or allowlists, sandboxing, restricted environment secrets, and egress monitoring. |
```
connor4312, line 76: >Require provenance, signatures or allowlists, sandboxing, restricted environment secrets, and egress monitoring

That's quite a high bar. I can't imagine many agent hosts will do this. We don't do any of this for plugins in VS Code today.

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+| **Denial of service and privacy leakage** | A peer sends huge JSON frames, deep state snapshots, large resources, unbounded terminal output, or logs prompts/tokens/file contents. | Enforce message, resource, history, subscription, and terminal scrollback limits. Apply backpressure and rate limits. Redact tokens, prompts, file contents, terminal output, and secrets from logs/telemetry. |
```
connor4312, line 77: A peer is already going to be authenticated. Standard best practice for resource management is good but I don't think this is a big deal. A client isn't authoritative on the state so I'm unsure what "Enforce message, resource, history, subscription, and terminal scrollback limits" would mean in that context

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+
+## Minimum requirements for implementations
+
+1. **Authenticate before protocol use.** Remote transports must authenticate peers before `initialize`; `clientId` must be bound to the authenticated connection.
```
connor4312, line 81: We don't do this. We let people stand up agent hosts and connect to them via websocket -- I have an agent host like this on my home server that I connect to from my devices. I don't tihnk we'd gain anything by removing this capability

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+2. **Encrypt non-local traffic.** Use `wss`, SSH, dev tunnel security, or equivalent protection for non-loopback connections.
+3. **Make trust local.** Clients and servers must make authorization decisions in their own policy layer, not from peer-provided text, labels, or confirmation flags.
+4. **Gate token delivery.** Sending OAuth/Bearer tokens to an agent host requires explicit consent and must be scoped to the intended resource and host.
+5. **Constrain client tools.** Do not expose powerful client tools to untrusted hosts by default. When enabled, require local allowlists, argument validation, and user confirmation for sensitive operations.
+6. **Constrain resource and terminal APIs.** Servers must sandbox filesystem access and terminal operations; clients that serve local resources must enforce their own scheme/path policy.
```
connor4312, line 86: >Servers must sandbox filesystem access and terminal operations

This requirement currently cannot be fulfilled on Windows, and even on other OS' I think is too heavy to have as a "minimum requirement". Sandboxing is complex.

THREAT_MODEL.md
```diff
@@ -0,0 +1,90 @@
...
+7. **Handle multi-client ownership.** Active-client state, tool calls, terminal claims, input requests, and approvals must be tied to the owning authenticated connection.
+8. **Treat plugins and packages as executable.** Customizations, MCP servers, package installs, hooks, and skills require supply-chain policy and sandboxing.
+9. **Validate and bound the protocol.** Use schema validation, fail closed on invalid messages, and enforce limits on frame size, JSON depth, resource size, subscriptions, replay history, and terminal output.
```
connor4312, line 89: Similar to above https://github.com/microsoft/agent-host-protocol/pull/88/changes#r3155649734

rwoll:
I'm going to close this and simplify including some minimal examples.

PR #93, opened 2026-05-01
swift sample: send messages via session/pendingMessageSet

colbylwilliams:
## What

Switch the `AHPClient` Swift sample's `AppStore.sendMessage` from dispatching `session/turnStarted` to dispatching `session/pendingMessageSet { kind: queued }`, and render queued/steering pending messages in the chat view.

## Why

`session/turnStarted` is technically marked `@clientDispatchable` in the protocol, but the docs scope the client-side use of that action narrowly to the **truncate-and-resume-with-edited-message** flow (see `docs/guide/state-model.md`).

The canonical "user sent a new message" flow is:

1. Client dispatches `session/pendingMessageSet { kind: queued, id, userMessage }`.
2. Per the spec — _"if the session is idle when a queued message is set, the server SHOULD immediately consume it and start a new turn"_ — the server emits `session/pendingMessageRemoved` followed by its own `session/turnStarted` (with `queuedMessageId` linking back).
3. If a turn is already in progress, the message naturally waits in the queue and is auto-started once the active turn completes.

The previous `session/turnStarted`-from-client approach worked in some hosts but doesn't match real session worker implementations (e.g. `copilot-host`), which only drive their underlying agent off the `pendingMessageSet` action. This PR aligns the sample with the documented contract.

## Notes

- Optimistically applies the `pendingMessageSet` locally so the message bubble shows up immediately as a "Queued" entry. The server-emitted `turnStarted` then promotes it into an active turn.
- Adds a small `PendingMessageView` that renders both `queuedMessages` and `steeringMessage` (faded user bubble + "Queued"/"Steering" caption) so multi-message and busy-session UX is visible to users.
- Auto-scroll fires when `queuedMessages.count` or `steeringMessage.id` changes.
- No protocol changes — this is sample-app-only.

## Verification

- `swift build` and `swift test` on the `AgentHostProtocol` Swift package both pass (14/14 tests).
- Built a small standalone target against the regenerated package to confirm the new `SessionPendingMessageSetAction(type:session:kind:id:userMessage:)`, `PendingMessage`, and `SessionState.{queuedMessages,steeringMessage}` shapes compile.
- The full iOS app build currently fails on `main` due to a separate, pre-existing protocol type drift in `InputRequestView.swift` and `RequestPartView.swift` (`MessageAttachment.path` → `uri`, `SessionInputRequest.message` becoming optional). Confirmed by stashing this PR's changes — the same errors reproduce. That cleanup is filed separately.


rebornix (approved):
awesome!

PR #99, opened 2026-05-02
Make Swift client consumable via Swift Package Manager

colbylwilliams:
## Summary

Make `AgentHostProtocol` consumable as a remote Swift Package Manager dependency:

```swift
.package(url: "https://github.com/microsoft/agent-host-protocol.git", from: "x.y.z")
```

SwiftPM only resolves manifests at the root of a remote git repo, so this PR moves the package manifest to the repository root while keeping the actual Swift sources where they live today (`clients/swift/AgentHostProtocol/`).

## Why

Today no external consumer can pull this in via SPM:
- The package's `Package.swift` lives in a subdirectory, which SwiftPM cannot resolve remotely.
- The bulk of the implementation (`Sources/AgentHostProtocol/Generated/`) was `.gitignore`d, so even if SPM could see the manifest there'd be nothing to compile.

## What changed

- **`.gitignore`** — Stop ignoring `Generated/`. Switch the per-subdir `.build/`/`.swiftpm/` ignores to root-level globs (the manifest now lives at the root).
- **`/Package.swift`** (new) — Hand-maintained at the repo root. Targets use `path:` to point into `clients/swift/AgentHostProtocol/{Sources,Tests}`.
- **`clients/swift/AgentHostProtocol/Package.swift`** — Deleted; replaced by the root manifest. (Two manifests for the same package would drift.)
- **`clients/swift/AHPClient/AHPClient.xcodeproj/project.pbxproj`** — `XCLocalSwiftPackageReference` retargeted from `../AgentHostProtocol` to `../../..` so the example app keeps consuming the same package via its new location.
- **`scripts/generate-swift.ts`** — No longer writes `Package.swift` (it's hand-maintained at the root now). Still owns `Generated/*.generated.swift` and seeds `AnyCodable.swift` if missing.
- **`clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/Generated/`** — Now committed. CI regenerates and fails on drift, so this can't go stale.
- **`ToolCallStateExtensions.swift`** — Added the missing `.systemNotification` case in the `partId` switch. Pre-existing exhaustiveness bug that only surfaced once `swift build` actually runs in CI.
- **`.github/workflows/ci.yml`** — New `swift` job on `macos-latest`: regenerates Swift, fails on diff, then runs `swift build` + `swift test`. Bumped common GitHub Actions to current majors across all three workflows: `actions/checkout` v4 → v6, `actions/setup-node` v4 → v6, `actions/upload-pages-artifact` v3 → v5, `actions/deploy-pages` v4 → v5. (`Swatinem/rust-cache@v2` is still the latest major.)
- **`README.md`, `clients/swift/AGENTS.md`** — Documented SPM consumption and the root-manifest layout.

## Design notes

- **Why not split the Swift package into its own repo?** Generated sources stay in lockstep with the TypeScript protocol definitions in the same commit; splitting would create a constant sync burden. The polyglot-monorepo with a root `Package.swift` is a small price.
- **Why a `Generated/` subfolder?** The package mixes hand-written reducers with generated types. The directory boundary makes it obvious to contributors which files get clobbered by `npm run generate:swift` and is a common Swift convention (SwiftGen, Sourcery, Apollo iOS, swift-protobuf).
- **The example AHPClient app is not declared in any SPM target**, so it never reaches a consumer's binary. SPM does still clone the whole repo on resolve (no sparse-checkout), but the AHPClient sources are ~320KB out of ~2.4MB tracked content — small enough that splitting the example into its own repo isn't worth the maintenance cost.

## Verification

- `swift build` ✅
- `swift test` ✅ 14/14 pass
- `npm test`, `npm run typecheck`, `npm run lint` ✅
- `npm run generate:swift` is idempotent against the committed tree ✅

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,9 +10,9 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@v6
         with:
           node-version: 24
           cache: npm
@@ -29,7 +29,7 @@ jobs:
       run:
         working-directory: clients/rust
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v6
 
       - uses: dtolnay/rust-toolchain@stable
         with:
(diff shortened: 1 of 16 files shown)


rebornix (approved):
Great, thanks for the help!

PR #106, opened 2026-05-05
First class text selection fields to AHP attachments

anthonykim1:
Motivated from: https://github.com/microsoft/vscode/pull/314245#discussion_r3190481067 
Seems like bunch of rust files and schema.json showing up due to npm run generate.


types/state.ts
```diff
@@ -909,6 +909,30 @@ export interface MessageAttachmentBase {
...
+ *
+ * @category Turn Types
+ */
+export interface TextDocumentRange {
+  /** Start position of the range. */
+  start: TextDocumentPosition;
+  /** End position of the range. */
+  end: TextDocumentPosition;
+}
+
```
connor4312, line 924: - I would name these "TextPosition" or "TextRange" -- AHP doesn't have a notion of a "text document" like vscode/lsp
- We should move from offset to ranges in the attachment anchors as well
- On thing that would be a pro to keeping offset-based logic is that it could work with binary attachment as well, allowing the user to reference byte ranges. But the semantics of bytes vs UTF-16 code points gets muddy. So maybe this is not worth persuing.

types/state.ts
```diff
@@ -945,6 +969,19 @@ export interface MessageEmbeddedResourceAttachment extends MessageAttachmentBase
...
+   * This is distinct from {@link MessageAttachmentBase.rangeStart} /
+   * {@link MessageAttachmentBase.rangeEnd}, which refer to the span in
+   * {@link UserMessage.text} that references the attachment.
+   */
+  documentRange?: TextDocumentRange;
+  /**
+   * Optional text covered by {@link documentRange}, when already known by the
+   * client or producer.
+   */
+  selectedText?: string;
```
connor4312: - Probably `documentRange` -> `range` to avoid the "document" vocabulary
- `selection?: { value: string, range: TextRange }` might be a little more compact and require them 
- currently these references can be used for binary data too. Do we saw that the selection only has meaning for textual resources?

types/state.ts
anthonykim1: Great, thanks! 🙂 @connor4312

* Renamed to TextPosition/TextRange --> [e5a4e80](https://github.com/microsoft/agent-host-protocol/pull/106/commits/e5a4e80cf3a555411e6f3e239978b1439c54d7f3) 
* Moved offset to ranges --> [d70a0b1](https://github.com/microsoft/agent-host-protocol/pull/106/commits/d70a0b15d55ad068bb03fbdd424663c0f01f9298) 
* For the binary case, I avoided defining byte ranges here and made the selection shape textual-only, since mixing byte offsets and text positions would make the semantics muddy like you mentioned.

If we need binary ranges later, should we add a separate byte-range shape with explicit byte semantics rather than overloading `TextRange`?

types/state.ts
anthonykim1: * Updated `documentRange` / `selectedText` to `selection?: { value, range }` --> [4088127](https://github.com/microsoft/agent-host-protocol/pull/106/commits/40881270d374090c238a207eb604ae151ede5eb1) 

* More updates to emphasize that selection from this PR is only meaningful for textual resource --> [a9c1740](https://github.com/microsoft/agent-host-protocol/pull/106/commits/a9c174057a7dcf2e6ff3829141cfd27db8c6aecb)

connor4312 (approved):
lgtm

types/state.ts
```diff
@@ -909,6 +898,46 @@ export interface MessageAttachmentBase {
...
+ *
+ * This is only meaningful for textual resources. Binary resources may still
+ * use resource or embedded resource attachments, but they should not use this
+ * text selection field.
+ *
+ * @category Turn Types
+ */
+export interface TextSelection {
+  /** The selected text. */
+  value: string;
```
roblourens, line 936: I'm not sure we should include the selected text inline. The document is normally provided by URI, the agent should resolve that then extract the text range

types/state.ts
```diff
@@ -909,6 +898,46 @@ export interface MessageAttachmentBase {
...
 }
 
+/**
+ * A zero-based position within a textual document.
+ *
+ * @category Turn Types
+ */
+export interface TextPosition {
+  /** Zero-based line number. */
+  line: number;
```
roblourens, line 908: The difficult thing with this is that when your content is just strings, not something like a document, then to resolve the line/character reference, the consumer has to count lines. Using offsets within the string might be a better fit?

types/state.ts
anthonykim1: Hmm..

My thinking is: 
* Required value is easiest for consumers, but forces producers to include text.
* No value is cleanest source-of-truth-wise, but forces consumers to resolve/slice resources...

Can't decide which one is better..

What if we make value optional? so then  producers like VS Code can provide exact text when they have it, consumers can ignore it and read the URI if they prefer.

types/state.ts
anthonykim1: Yeah, maybe we can better handle two cases  here: 
1) attachment anchors in `UserMessage.text` point into the plain chat message string, so they can use offset ranges
2) selections inside a textual resource point into the attached file/resource, so they can keep using `TextRange`


Something like: 
```
const userMessage: UserMessage = {
  text: 'Explain this @foo.ts',
  attachments: [
    {
      type: MessageAttachmentKind.Resource,
      uri: 'file:///repo/foo.ts',
      label: 'foo.ts',
      displayKind: 'selection',

      // Anchor range in UserMessage.text for "@foo.ts"
      rangeStart: 13,
      rangeEnd: 20,

      // Selection range inside the referenced textual resource
      selection?: {
        value: 'const selectedValue = 123;',
        textRange: {
          start: { line: 10, character: 4 },
          end: { line: 12, character: 8 },
        },
      },
    },
  ],
};
```

types/state.ts
anthonykim1: @roblourens @connor4312 Made: https://github.com/microsoft/agent-host-protocol/pull/107 wanting to address feedback!

types/state.ts
anthonykim1: Also: https://github.com/microsoft/agent-host-protocol/pull/107

PR #107, opened 2026-05-06
Remove value from TextSelection

anthonykim1:
Follow-up to: https://github.com/microsoft/agent-host-protocol/pull/106 
Wanting to address comments from discussions: 
* https://github.com/microsoft/agent-host-protocol/pull/106#discussion_r3191987272 
* https://github.com/microsoft/agent-host-protocol/pull/106#discussion_r3191990354

### DIFF
--- a/types/state.ts
+++ b/types/state.ts
@@ -943,8 +943,6 @@ export interface TextRange {
  * @category Turn Types
  */
 export interface TextSelection {
-  /** The selected text, when already known by the producer. */
-  value?: string;
   /** The range covered by the selection. */
   range: TextRange;
 }
--- a/types/state.ts
+++ b/types/state.ts
@@ -842,9 +842,9 @@ export interface ActiveTurn {
  * A user message and its associated attachments.
  *
  * Attachments MAY be referenced inside {@link UserMessage.text} via their
- * {@link MessageAttachmentBase.rangeStart}/{@link MessageAttachmentBase.rangeEnd}
- * fields. Attachments without a range are still associated with the message
- * but do not correspond to a specific span in the text.
+ * {@link MessageAttachmentBase.range} field. Attachments without a range are
+ * still associated with the message but do not correspond to a specific span
+ * in the text.
  *
  * @category Turn Types
  */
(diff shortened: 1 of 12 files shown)


types/state.ts
```diff
@@ -932,9 +943,9 @@ export interface TextRange {
  * @category Turn Types
  */
 export interface TextSelection {
-  /** The selected text. */
-  value: string;
-  /** The range covered by {@link value}. */
+  /** The selected text, when already known by the producer. */
+  value?: string;
```
roblourens: I'm just not sure we should have it at all. It can massively blow up the size of messages, it's inconsistent with the rest of the protocol, and is redundant if you provide a URI and a range

connor4312: I think the value is useful specifically in the case of the resource being provided as a `ContentRef`. The file might change in the meantime which would garble the selection as initially given (not hard to hit if you have an agent working while you compose your next message.) Including the `value` avoids this issue, but it is superfluous if given as a `MessageEmbeddedResourceAttachment`

types/state.ts
```diff
@@ -868,10 +868,21 @@ export interface MessageAttachmentBase {
...
+   * If defined, the start of the range in {@link UserMessage.text} that
+   * references this attachment. The range is the half-open interval
+   * `[rangeStart, rangeEnd)` of character offsets, measured in UTF-16 code
+   * units.
+   *
+   * When present, `rangeEnd` MUST also be present and MUST be greater than or
+   * equal to `rangeStart`.
    */
-  range?: TextRange;
+  rangeStart?: number;
```
connor4312: My thinking with pushing this direction is just for coherence with DAP and LSP. I initially did offset-based selections because I didn't want to bring in a notion of positions _just_ for completions, but now that we have them I think that we should consistently represent text ranges as either offsets or line/col, and line/col matches other protocols.

types/state.ts
roblourens: But this is also unbounded inline content which will be broadcast to all clients whether they need it or not. I might cmd+A in a huge file and send a ton of content right?

connor4312: True. And I could cmd+a a 100mb file, in theory. So maybe we leave this out until consumers ask for it. I think we should just get rid of the property entirely in that case.

I would leave TextSelection as its own interface though so we have the option to add it back in a non-breaking way if we want

dmitrivMS (dismissed):

types/state.ts
anthonykim1: Makes sense, thank you both!
I'll get rid of selection for now, but leave TextSelection as its own interface.

types/state.ts
anthonykim1: Make sense, brought them back to stay consistent with other protocols :)

roblourens (approved):

PR #110, opened 2026-05-07
feat: Introduce startTurn command to replace session/turnStarted action

DonJayamanne:
- Added `startTurn` command to initiate a new turn in an existing session, allowing server validation of session configuration.
- Deprecated the legacy `session/turnStarted` action in favor of the new command.
- Updated session configuration handling to support dynamic schema updates and client-dispatched configuration changes.
- Enhanced error handling for invalid session configurations during turn starts.
- Updated related schemas, types, and reducer logic to accommodate the new command and configuration management.
- Added test cases to ensure correct behavior of session configuration changes and turn initiation.


types/commands.ts
```diff
@@ -1058,3 +1067,91 @@ export interface CompletionsResult {
...
+  /** Client-chosen turn identifier */
+  turnId: string;
+  /** User's message */
+  userMessage: UserMessage;
+  /**
+   * If this turn is auto-started from a queued message, the ID of that
+   * message. Mirrors the same field on the legacy
+   * {@link import('./actions.js').SessionTurnStartedAction}.
+   */
+  queuedMessageId?: string;
```
connor4312: Queued messages are processed by the agent host, there shouldn't be a scenario where a client would dequeue this I think

types/commands.ts
```diff
@@ -1058,3 +1067,91 @@ export interface CompletionsResult {
...
+}
+
+/**
+ * Result of the `startTurn` command. `null` on success; failures surface as
+ * JSON-RPC errors (see {@link StartTurnParams}).
+ */
+export type StartTurnResult = null;
```
connor4312: We've been using an empty object for this

types/commands.ts
```diff
@@ -1058,3 +1067,91 @@ export interface CompletionsResult {
...
+ * when the session's config does not satisfy the current schema. Helps
+ * clients route the user back to the offending fields in the config picker.
+ *
+ * @category Errors
+ */
+export interface StartTurnInvalidConfigErrorData {
+  /** Required schema property ids that are missing from `state.config.values`. */
+  missingRequired?: string[];
+  /**
+   * Property ids whose values failed schema validation, with a short
```
connor4312: Are there errors that are expected here? An error like wrong-type or not-in-enum would be a logic error on the client that the server can just reject with InvalidParams

types/actions.ts
```diff
@@ -230,6 +231,10 @@ export interface SessionCreationFailedAction {
  * @category Session Actions
  * @version 1
  * @clientDispatchable
+ * @deprecated Prefer the rejectable `startTurn` command (since v0.1.0
+ * additive). The action remains client-dispatchable for backwards
+ * compatibility; the server still emits this action to subscribers when a
+ * turn begins, regardless of which path triggered it.
```
connor4312: I don't think we need a particular note here -- no public consumers yet

roblourens:
I wonder whether we could get rid of the session config changing at all, to simplify further. 

- create a session for a working directory
- the working directory never changes. The one truly dynamic piece of the config is options that are based on whether the cwd is a git repo. That shouldn't change so it can be decided once for the session.
- The other part- deciding whether 'branch' is enabled, based on whether you're using worktree mode- we could express with a more complicated json schema using `if` or `oneOf`. That would be more complex for the client rendering a form but simplify the protocol. It's hard for me to come up with legit reasons for the config to be fully dynamic.

Issue #92, opened 2026-04-30
RFC: changesets


connor4312:
We want to be able to represent different kinds of file changes in the protocol. For this I'm proposing a new notion of "changesets":

---

# Changesets

A **changeset** is a named, individually subscribable view of file changes associated with a session. Changesets replace the legacy `summary.diffs` field and generalise it: a session can expose any number of changesets — uncommitted working-tree edits, the diff between two turns, the cumulative changes for the whole session, the staged index, etc. — each with its own URI, lifecycle, and update stream.

This doc proposes the design. It is **draft** — not yet implemented.

## Why

Today, AHP carries a single `FileEdit[]` on `SessionSummary.diffs` that the server replaces wholesale with `session/diffsChanged`. That model has obvious limits:

- **Only one diff per session.** No way to express "uncommitted vs HEAD" alongside "this turn's edits" alongside "everything since session start".
- **No subscription granularity.** Clients receive the full list on every change, even if they only care about one slice.
- **No actions.** A client cannot stage, revert, or open a PR from a diff without reaching outside the protocol.
- **No turn-relative diffs.** A user may want to compare the working tree at the start of turn `t1` against turn `t3`, but the protocol has no shape for that.

Changesets address all four.

### Changeset Catalogue

Each session's `SessionSummary` advertises the set of changesets the server can produce. The summary entry is intentionally lightweight — just enough to render a chip or list row without subscribing — and references a full subscribable `ChangesetState` by URI.

> Note: session summaries are what's exposed on the base `listSessions()` result. So a client who's loading a session can request both the changeset and session details concurrently. This should let the default changeset load quickly -- it's a single round trip to the agent host.

```typescript
SessionSummary {
  // ...existing fields, with `diffs` removed...
  changesets?: ChangesetSummary[]
}

ChangesetSummary {
  /** Stable identifier, unique within the session. */
  id: string
  /** Human-readable label, e.g. `"Uncommitted Changes"`. */
  label: string
  /**
   * RFC 6570 URI template. Clients parse the variables directly out of the
   * template using the standard `{name}` syntax — they are not redeclared
   * here.
   */
  uriTemplate: string
  /** Optional longer description. */
  description?: string
  /** Aggregate line additions across the changeset, when known. */
  additions?: number
  /** Aggregate line deletions across the changeset, when known. */
  deletions?: number
  /** Number of files in the changeset, when known. */
  files?: number
}
```

[...]

PR #116, opened 2026-05-13
feat(changesets): add changesets for sessions

DonJayamanne:
For https://github.com/microsoft/agent-host-protocol/issues/92

- Introduced a new file `v2.ts` containing frozen wire-format types for protocol version 2.
- Ensured bidirectional assignability between the frozen types and current living types.
- Captured the shape of every state type and action type at the time of version 2 release.
- Established compatibility assertions to prevent breaking changes in future updates.


types/state.ts
```diff
@@ -402,8 +402,158 @@ export interface SessionSummary {
...
+  /** Stable identifier, unique within this changeset. */
+  id: string;
+  /** Human-readable button/menu label. */
+  label: string;
+  /** Optional longer description shown on hover or in tooltips. */
+  description?: string;
+  /** Where this operation can be invoked. */
+  scopes: ChangesetOperationScope[];
+  /** Hint for clients to confirm before invoking. */
+  destructive?: boolean;
```
connor4312: Maybe a `confirmation?: string` would be better instead? Otherwise a client could only have a generic message like "Are you sure you want to do this?"

types/state.ts
```diff
@@ -402,8 +402,158 @@ export interface SessionSummary {
...
+  /**
+   * RFC 6570 URI template. Clients parse the variables directly out of the
+   * template using the standard `{name}` syntax — they are not redeclared
+   * here.
+   *
+   * Only the variable names `turnId`, `originalTurnId`, and
+   * `modifiedTurnId` are defined by this protocol. Templates with no
+   * variables are themselves subscribable URIs. Clients SHOULD ignore
+   * templates containing unknown variable names — there is no
+   * protocol-defined way to obtain values for them.
```
connor4312: Let's include the full semantics included in the original issue (a single turnId vs original/modifiedTurnId vs no turn ID)

types/reducers.ts
```diff
@@ -584,12 +584,6 @@ export function sessionReducer(state: SessionState, action: SessionAction, log?:
         summary: { ...state.summary, activity: action.activity },
       };
 
-    case ActionType.SessionDiffsChanged:
```
connor4312, line 587: Seem to be missing any new action to update changeset states

types/state.ts
```diff
@@ -402,8 +402,158 @@ export interface SessionSummary {
...
+ * Catalogue entries are intentionally lightweight — just enough to render a
+ * chip or list row without subscribing. Full per-changeset detail
+ * ({@link ChangesetState}) lives on the subscribable URI obtained by
+ * expanding {@link uriTemplate}.
+ *
+ * @category Changesets
+ */
+export interface ChangesetSummary {
+  /** Stable identifier, unique within the session. */
+  id: string;
```
connor4312: I don't think we need an ID here -- the uriTemplate should be unique. That's what the AH is going to need to use to resolve the changeset, after all.

types/commands.ts
```diff
@@ -1058,3 +1058,78 @@ export interface CompletionsResult {
...
+ * Identifies the file or range a {@link ChangesetOperation} should act on.
+ *
+ * The `kind` MUST match one of the operation's declared
+ * {@link ChangesetOperation.scopes}.
+ *
+ * @category Commands
+ */
+export type ChangesetOperationTarget =
+  | { kind: 'resource'; resource: URI; side?: 'before' | 'after' }
+  | { kind: 'range'; resource: URI; side?: 'before' | 'after'; range: { start: number; end: number } };
```
connor4312: Now that we have `TextRange` in the protocol, we should use that

types/commands.ts
```diff
@@ -1058,3 +1058,78 @@ export interface CompletionsResult {
...
+/**
+ * Optional follow-up surfaced by the server after an operation completes —
+ * typically a URI to open (e.g. a created PR) or a {@link ContentRef} the
+ * client can fetch.
+ *
+ * @category Commands
+ */
+export type ChangesetOperationFollowUp =
+  | { kind: 'openUri'; uri: URI }
+  | { kind: 'content'; content: ContentRef };
```
connor4312: - The `kind` in both of these should be enums to match our existing patterns
- I wonder if we need two ChangesetOperationFollowUp. ContentRef is already a URI. Maybe we have one followup kind (at least for now) and have `external?: boolean`

Yoyokrazy (dismissed):

types/state.ts
```diff
@@ -402,8 +402,158 @@ export interface SessionSummary {
...
+   */
+  operations?: ChangesetOperation[];
+}
+
+/**
+ * One file entry within a {@link ChangesetState}.
+ *
+ * @category Changesets
+ */
+export interface ChangesetFile {
```
roblourens, line 506: I guess I'm wondering why this exists, rather than just use FileEdit? Why does it have its own id?

types/state.ts
```diff
@@ -402,8 +402,158 @@ export interface SessionSummary {
...
+ * A server-declared invokable verb the client can run against a
+ * changeset, a file, or a range — `"stage"`, `"revert"`, `"create-pr"`,
+ * and so on.
+ *
+ * The term "operation" is used deliberately to avoid colliding with the
+ * protocol-level [Actions](/guide/actions) that mutate state.
+ *
+ * @category Changesets
+ */
+export interface ChangesetOperation {
```
roblourens, line 548: `revert` makes sense to me. But bringing in "staging" when we don't have the concept of showing that a file is staged, seems odd. And I'm unsure about 'create-pr' as an agenthost operation. I guess this feels like it's just halfway to implementing full SCM support but is in a weird place?

types/commands.ts
roblourens: Yeah this is another point that I'm not sure we can have a good experience for something like create-pr as a totally generic action.

types/state.ts
DonJayamanne: Fixed

types/state.ts
DonJayamanne: Fixed

types/reducers.ts
DonJayamanne: Fixed

types/state.ts
DonJayamanne: Yeah, this is interesting.
Perhaps we'll need to review the code in VS code for this.

For sessions we need to store the summary of the session changes in SQL `changeset/session`
Currently I'm not hardcoding the URI for that, i'm using a stable hardcoded id of `session`
This way I can use that everywhere in VS Code to look for this particular item
This is the item that will always be on top in the changesets and will always be persisted as well

Without the id, i'd have to revert to having a function to hardcode the URI, that is also correct.
However I believe this may highlight a problem in my implementation
Lets discuss the code in VS Code tomorrow, please ping offline when you're free.

types/commands.ts
DonJayamanne: Fixed

types/state.ts
DonJayamanne: Fixed

types/state.ts
DonJayamanne: >   /** Stable identifier within the changeset. Typically `after.uri` (or `before.uri` for deletions). */
>  id: string

This is crucial for solving renames.
The client will not be able to figure out when a change comes in, which file it is for, 
Rename A=>B, and then B=>C and then create a new file A or B, these all could lead to a few issues with how host and client maps the changes. both need to agree on what is sent in before and after. 
I can see this failing, i believe thats why Connor added this as well.

>  /** Server-defined opaque metadata, surfaced to operations and tooling but not interpreted by the protocol. > */
>  _meta?: Record<string, unknown>

It also carries the `_meta` field which is part of the original requirement.
Its for future proofing with some level of extensibility.

types/commands.ts
DonJayamanne: Fixed

types/state.ts
connor4312: Yea it's just a bit fuzzier. To handle renames, deletions, and creations one would use `edit.after.uri || edit.before.uri` to get a unique identifying URI. This is just a bit cludgy to have consumers to every time.

types/state.ts
connor4312: As discussed it sounds like we can drop the `id` here

types/commands.ts
```diff
@@ -1058,3 +1058,110 @@ export interface CompletionsResult {
...
+ * - `OpenUri` — open in the user's preferred external handler (e.g. browser).
+ *
+ * @category Commands
+ */
+export interface ChangesetOperationFollowUp {
+  kind: ChangesetOperationFollowUpKind;
+  content: ContentRef;
+  /** When `true`, open in an external handler rather than inline. */
+  external?: boolean;
+}
```
connor4312: `kind` vs `external` are duplicative. I would drop `kind` or just have it be a single enum action for now, and keep `external`

types/commands.ts
```diff
@@ -1058,3 +1058,110 @@ export interface CompletionsResult {
...
+ * command.
+ *
+ * `ok: true` indicates the server accepted the operation; the operation
+ * MAY still produce subsequent failure feedback through the
+ * {@link ChangesetStatusChangedAction | `changeset/statusChanged`} stream.
+ *
+ * @category Commands
+ */
+export interface InvokeChangesetOperationResult {
+  ok: boolean;
```
connor4312: I'd think the host should reject the call with an error instead of setting `ok: true`

pwang347 (dismissed):

connor4312 (dismissed):

aeschli (approved):

PR #120, opened 2026-05-14
refactor: update CompletionItem attachment handling for text-expansion completions

DonJayamanne:


connor4312:
as discussed, I think we still want attachments here because they mark ranges in the text. We might just have a new advisory of `displayKind` being a 'command' or 'skill'

connor4312:
think we're good to close now, lmk if not

PR #121, opened 2026-05-14
Add ahp::hosts multi-host SDK to the Rust client

colbylwilliams:
## Why

A consumer that wants to talk to two or more AHP hosts at once today has to hand-roll all of: N independent `Client`s, N transports, N reconnect supervisors, a per-host metadata registry, a fan-in of inbound events tagged with which host produced them, per-host scoping of resource URIs (`copilot:/s1` on Host A ≠ `copilot:/s1` on Host B), persistence of `clientId` per host so reconnect identity survives restarts, and a per-host root state mirror + session summary cache (otherwise sidebars/inboxes degrade to "subscribe to everything"). This is a real common-case need for any UI consumer (Tauri/desktop app, iOS, Android) that lets users register more than one agent host.

## What

Adds a `MultiHostClient` to the Rust SDK in a new `ahp::hosts` module. Designed so single-host = N=1 of multi-host (`MultiHostClient::single(config)` is a one-line constructor), with no separate single-host API to learn. Per-host UX details (label, transport, connection state, last error, protocol version, agents, `lastSeenServerSeq`, subscriptions, session summaries, etc.) are first-class on `HostHandle`. Aggregated cross-host views are real types backed by a real cache, not bare tuples.

### Per-host runtime

- `HostRuntime` task per host owns the current `Client`, the reconnect state machine (`ReconnectPolicy` with backoff + jitter), the per-host root-state mirror, and a session-summary cache (seeded by `listSessions`, kept fresh by `notify/sessionAdded`/`Removed`/`SummaryChanged`).
- Every (re)connect bumps a generation counter. `HostClientHandle`s refuse to dispatch through a stale connection, returning `HostError::HostReconnected` — this prevents subtle bugs where a handle held across a reconnect silently writes to a different connection.
- Stable `clientId` per host through a pluggable `ClientIdStore` (default `InMemoryClientIdStore`; consumers should plug a persistent backend for cross-launch reconnect identity).

### Public surface (`ahp::hosts`)

- `HostId`, `HostConfig`, `HostState`, `HostHandle`, `HostClientHandle`, `HostError`, `HostEvent`, `HostSubscriptionEvent`, `HostedSessionSummary`, `HostedAgent`.
- `HostTransportFactory` — blanket impl for any `Fn(HostId) -> Future<Output = Result<BoxedTransport, _>>`, so the common case is a closure.
- `MultiHostClient`: `new` / `single` / `with_client_id_store` / `add_host` / `remove_host` / `reconnect_host` / `host` / `hosts` / `client` / `subscribe` / `unsubscribe` / `dispatch` / `events` / `host_events` / `aggregated_sessions` / `aggregated_agents`.
- `events()` and `host_events()` return fresh multicast receivers per call. Ordering is **per host only**; there is no cross-host total order. Slow consumers skip the gap (matching `SessionSubscription` semantics).

### Supporting changes

- New `ahp::transport::BoxedTransport` + `DynTransport` adapter so the multi-host runtime can store heterogeneous transports per host (the existing `Transport` trait stays generic and zero-cost).
- New `Client::events()` exposes a top-level fan-out of every inbound `SubscriptionEvent` tagged with its resource URI (`Some(uri)` for actions, `None` for protocol notifications). The runtime attaches a receiver **before** the `initialize` handshake so notifications pushed during the handshake aren't dropped.
- New `ahp_types::ROOT_RESOURCE_URI` constant. The Rust SDK now consistently uses `agenthost:/root` (matching the spec, docs, and Swift example) instead of the stale `root:/` fallback that was in `Client::action_resource()` and the example/READMEs. This was a pre-req for correct fan-out of root-state actions through the multi-host layer.

## Tests + docs

- `crates/ahp/tests/hosts.rs`: 8 integration tests using an in-memory transport pair (mirroring `client_roundtrip.rs`):
  - `single_constructor_yields_connected_handle`
  - `two_hosts_register_and_connect_independently`
  - `aggregated_sessions_track_listsessions_then_notification`
  - `host_client_handle_invalidates_after_reconnect`
  - `remove_host_terminates_supervisor_and_emits_event`
  - `fan_in_events_carry_host_id_and_resource`
  - `transport_factory_is_called_for_each_reconnect`
  - `duplicate_host_id_is_rejected`
- `crates/ahp/src/hosts/policy.rs`: 5 unit tests for backoff/jitter math.
- 3 new doctests in the `ahp::hosts` module.
- New `docs/guide/clients-multi-host.md` covering the design and surface, linked from the docs sidebar (Guide → Clients → Multiple Hosts) and from the root + per-crate READMEs.

## Compatibility

No protocol/wire changes. No CodeGen impact. All existing Rust tests still pass; the `ahp::hosts` module and `Client::events()` / `BoxedTransport` are pure-additive — no breaking changes.

## Out of scope (follow-ups)

The same shape needs to land in Swift: a new `AgentHostProtocolClient` SwiftPM product (single-host `AHPClient` + `AHPTransport` protocol, extracted/cleaned from the example app's `AHPConnection` patterns), then a Swift `MultiHostClient` mirroring the Rust API. Refactoring the example `AHPClient` iOS app onto `MultiHostClient` is a third, validation-only PR. These are intentionally split so SDK design changes aren't buried under either Swift onboarding or app churn.


clients/rust/crates/ahp/src/hosts/types.rs
```diff
@@ -0,0 +1,516 @@
...
+        /// One-based attempt counter. Resets to 1 after a successful connection
+        /// when [`ReconnectPolicy::reset_on_success`] is `true`.
+        attempt: u32,
+    },
+    /// Reconnect attempts were exhausted (or [`ReconnectPolicy::disabled`]
+    /// was configured) and the host is no longer trying.
+    Failed {
+        /// Human-readable reason. Mirrors [`HostHandle::last_error`].
+        reason: String,
+    },
```
connor4312, line 77: We should probably have an actual Error type here, feel free to grab thiserror

clients/rust/crates/ahp/src/hosts/factory.rs
```diff
@@ -0,0 +1,122 @@
...
+/// store returns `Some`, that id is reused — letting the server treat
+/// successive launches as the same client (which the AHP `reconnect`
+/// flow needs to replay missed actions). If the store returns `None`,
+/// the multi-host client generates a fresh UUID and stores it.
+///
+/// The default implementation is [`InMemoryClientIdStore`], which is
+/// session-stable but does **not** survive process restarts. Plug a
+/// keychain/file-backed implementation in for production multi-host
+/// apps so reconnects keep working across launches.
+pub trait ClientIdStore: Send + Sync + 'static {
```
connor4312: Clients can probably just expose an ID string and store it themselves as appropriate. I don't think there's much need to have a host ID -> client ID mapping, they can be reused.

clients/rust/crates/ahp/src/hosts/multi.rs
```diff
@@ -0,0 +1,404 @@
...
+            guard.get(id).map(|e| HostHandleTxRef {
+                cmd_tx: e.cmd_tx.clone(),
+                shared: e.shared.clone(),
+            })
+        }?;
+        entry.client_handle().await
+    }
+
+    /// Convenience: subscribe to `uri` on `host_id`.
+    pub async fn subscribe(
```
connor4312, line 227: Should subscribe (here or in general) return a type that `impl Drop` such that subscriptions are refcounted by the SDK?

clients/rust/crates/ahp/src/hosts/types.rs
```diff
@@ -0,0 +1,516 @@
...
+                Err(broadcast::error::RecvError::Closed) => return None,
+                Err(broadcast::error::RecvError::Lagged(_)) => continue,
+            }
+        }
+    }
+}
+
+/// Connection-level event for UX.
+#[derive(Debug, Clone)]
+pub enum HostEvent {
```
connor4312, line 405: non_exhaustive

connor4312 (commented):
some light review

clients/rust/crates/ahp/src/hosts/types.rs
colbylwilliams: @connor4312 did you mean to link something to `thiserror`?

connor4312: https://docs.rs/thiserror/latest/thiserror/ is a hand library for error macros

clients/rust/crates/ahp/src/hosts/types.rs
colbylwilliams: Done — added `#[non_exhaustive]` to `HostState`, `HostEvent`, and `HostError` so we can grow them without breaking changes. Fixed in fe3c250.

clients/rust/crates/ahp/src/hosts/types.rs
colbylwilliams: Done — `HostState::Failed.error` and `HostHandle.last_error` are now `Arc<ClientError>` so the typed error survives, and the docs explain that `HostState` skips `PartialEq` on `Failed` because `serde_json::Error` isn't `PartialEq`. Fixed in fe3c250.

clients/rust/crates/ahp/src/hosts/factory.rs
colbylwilliams: Done — removed `ClientIdStore` / `InMemoryClientIdStore` entirely. `HostConfig::new` now generates a session-stable UUID by default, and consumers persist the value themselves (e.g. via keychain) and pass it on subsequent launches via `HostConfig::with_client_id`. Also obsoletes Copilot's separate concern about `ClientIdStore::store` not being able to report persistence failures. Fixed in fe3c250.

clients/rust/crates/ahp/src/hosts/multi.rs
colbylwilliams: Good question — leaving as-is for now. RAII subscription handles with refcounting are a real API improvement, but they're also a substantive shape change (per-URI refcount table, `Drop` wakes the unsubscribe), and I'd want to land it across both the Rust and Swift SDKs together so they stay in sync. Happy to file a follow-up issue tracking it for both SDKs if you're up for that — let me know.

clients/rust/crates/ahp/src/transport.rs
```diff
@@ -35,8 +35,34 @@
...
+//!
+//! [`Transport`] is intentionally generic and not object-safe — the
+//! `impl Future` returns let it stay zero-cost on the hot path. When you
+//! need to store transports of different concrete types behind one
+//! handle (a registry of hosts, a transport factory that picks WebSocket
+//! vs stdio at runtime, etc.), wrap each one in [`BoxedTransport`]:
+//!
+//! ```no_run
+//! # async fn open_a() -> Result<impl ahp::Transport, ahp::TransportError> { unimplemented!() as Result<ahp::transport::BoxedTransport, _> }
+//! # async fn open_b() -> Result<impl ahp::Transport, ahp::TransportError> { unimplemented!() as Result<ahp::transport::BoxedTransport, _> }
```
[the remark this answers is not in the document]
colbylwilliams: Done — fixed in fe3c250 the way you suggested. The hidden helpers now return concrete `Result<BoxedTransport, TransportError>` and `unimplemented!()` coerces directly; the awkward `as Result<...>` cast is gone.

docs/guide/clients-multi-host.md
```diff
@@ -0,0 +1,142 @@
...
+# Ok(()) }
+```
+
+Configuration knobs live on `HostConfig` (`with_client_id`, `with_initial_subscriptions`, `with_client_config`, `with_reconnect_policy`) and on `ReconnectPolicy::{disabled, immediate_forever, exponential}`. For persistent identity across launches, build the multi-host client with a custom store via `MultiHostClient::with_client_id_store(...)`.
+
+See the `ahp::hosts` rustdoc and `crates/ahp/tests/hosts.rs` for the full surface.
+
+## Swift
+
+The Swift multi-host client lives in the `AgentHostProtocolClient` SwiftPM product and mirrors the Rust shape: `HostId`, `HostConfig`, `HostHandle`, `HostClientHandle`, `MultiHostClient`, `MultiHostClient.single(...)`, `events()`/`hostEvents()`, `aggregatedSessions`/`aggregatedAgents`. See the [Swift client guide](#) for usage. _(Coming with the Swift PRs.)_
```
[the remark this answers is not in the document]
colbylwilliams: Done — replaced the Swift section with a forward-looking note: "A matching `MultiHostClient` is planned for the Swift SDK in a follow-up PR... This page will be updated when the Swift implementation lands." Fixed in fe3c250.

clients/rust/crates/ahp/src/hosts/multi.rs
```diff
@@ -0,0 +1,404 @@
...
+    /// `Connecting`, or already `Reconnecting` if the first attempt
+    /// failed.
+    ///
+    /// Returns [`ClientError::Shutdown`] if the host id is already in
+    /// use; remove the existing host first.
+    pub async fn add_host(&self, config: HostConfig) -> Result<HostHandle, ClientError> {
+        let id = config.id.clone();
+        let mut hosts = self.inner.hosts.write().await;
+        if hosts.contains_key(&id) {
+            return Err(ClientError::Shutdown);
```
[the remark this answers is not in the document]
colbylwilliams: Done — `add_host` now returns `HostError::DuplicateHost(HostId)` (also new), so the failure mode is distinct from `ClientError::Shutdown`. Test renamed to `duplicate_host_id_is_rejected_with_typed_error`. Fixed in fe3c250.

README.md
```diff
@@ -13,6 +13,8 @@ The Agent Host Protocol (AHP) defines how a portable, standalone sessions server
 - **Swift** — Add `https://github.com/microsoft/agent-host-protocol` as a Swift Package Manager dependency to use the `AgentHostProtocol` library. See [`clients/swift/`](clients/swift/) for an example iOS client. The `Package.swift` manifest lives at the repository root because SwiftPM only resolves manifests at the root of a remote git repo; the actual Swift sources live under `clients/swift/AgentHostProtocol/`.
 - **Rust** — See [`clients/rust/`](clients/rust/) for the `ahp`, `ahp-types`, and `ahp-ws` crates.
 
+For consumers that need to talk to two or more hosts at once, both SDKs ship a `MultiHostClient` abstraction. See [Connecting to Multiple Hosts](https://microsoft.github.io/agent-host-protocol/guide/clients-multi-host) for the full surface; single-host consumers use the same API via `MultiHostClient::single`.
```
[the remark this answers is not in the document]
colbylwilliams: Done — softened to "the Rust SDK ships a `MultiHostClient`... A matching Swift API is planned in a follow-up PR." (For context: the Swift implementation is up as PRs #122 and #124, but neither is merged yet, so this PR's docs shouldn't claim Swift consumers can use the API today.) Fixed in fe3c250.

clients/rust/crates/ahp/src/hosts/factory.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Look up the previously stored `clientId` for `host_id`, if any.
+    fn load(&self, host_id: &HostId) -> Option<String>;
+
+    /// Persist `client_id` for `host_id`.
+    ///
+    /// Implementations should overwrite any previous value.
+    fn store(&self, host_id: &HostId, client_id: &str);
```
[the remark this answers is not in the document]
colbylwilliams: Moot now — per @connor4312's feedback above, `ClientIdStore` is removed entirely. `HostConfig::new` generates a session-stable UUID by default; consumers persist the value themselves (e.g. via keychain) and pass it back via `HostConfig::with_client_id` on subsequent launches. Persistence failures are now the consumer's responsibility to surface, with the full error type they choose. Fixed in fe3c250.

clients/rust/crates/ahp/src/client.rs
```diff
@@ -130,6 +175,7 @@ type PendingMap = HashMap<u64, oneshot::Sender<Result<Value, JsonRpcError>>>;
 struct Shared {
     pending: Mutex<PendingMap>,
     subscriptions: Mutex<HashMap<String, broadcast::Sender<SubscriptionEvent>>>,
+    all_events: broadcast::Sender<ClientEvent>,
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — fixed in fe3c250. `Shared.all_events` is now `std::sync::Mutex<Option<broadcast::Sender<ClientEvent>>>` and `drive_transport` drops the sender during teardown so `events.recv()` resolves to `None` after the underlying transport closes. Regression test: `client_events_recv_returns_none_after_transport_close`.

clients/rust/crates/ahp/src/hosts/runtime.rs
```diff
@@ -0,0 +1,734 @@
...
+            .await
+            .ok()?;
+        rx.await.ok()
+    }
+
+    pub(super) async fn shutdown(self) {
+        let _ = self.cmd_tx.send(HostCommand::Shutdown).await;
+        // Drop the sender first so the runtime sees `None` if Shutdown
+        // wasn't delivered before the receive loop exited.
+        drop(self.cmd_tx);
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — fixed in fe3c250. The supervisor now races `connect_once` (and the backoff sleep) against a per-host `Arc<Notify>` that `shutdown` triggers via `notify_one()` (queues a permit so the signal isn't lost if the supervisor isn't yet awaiting on it). A hung transport factory no longer blocks `remove_host`. Regression test: `shutdown_is_not_blocked_by_a_hung_transport_factory` — uses a factory that sleeps for an hour and asserts `remove_host` returns within 2 seconds.

clients/rust/crates/ahp/src/hosts/runtime.rs
```diff
@@ -0,0 +1,734 @@
...
+            if can_reconnect {
+                match client
+                    .reconnect(self.client_id.clone(), server_seq, subscriptions.clone())
+                    .await
+                {
+                    Ok(_result) => {
+                        // Reconnect succeeds; keep the existing serverSeq for now —
+                        // replayed actions and snapshots will catch us up. We
+                        // optimistically use the server's reconciliation here.
+                        (server_seq, None, true)
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — this was the same bug we'd separately filed as #125. Fixed in fe3c250: `apply_reconnect_result` now (a) for `Replay`, fans every replayed envelope through the per-host event tap and state mirror in serverSeq order and drops `missing` URIs from the local subscription set; (b) for `Snapshot`, refreshes the root-state mirror, advances `server_seq`, and drops URIs the server didn't snapshot. Regression test: `reconnect_replay_actions_are_fanned_out_with_advanced_seq`.

clients/rust/crates/ahp/src/hosts/multi.rs
```diff
@@ -0,0 +1,404 @@
...
+        }
+        let tx = spawn(
+            config,
+            self.inner.client_id_store.clone(),
+            self.inner.fan_out.clone(),
+            self.inner.host_events.clone(),
+        );
+        let snapshot = tx.snapshot().await.ok_or(ClientError::Shutdown)?;
+        hosts.insert(id, tx);
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — fixed in fe3c250 by taking the alternative you suggested. `add_host` now reserves the host id under the registry write lock, releases the lock, and reads the initial snapshot directly from per-host shared state instead of round-tripping through the supervisor's command channel. Same fix applied to `host`, `hosts`, and `client` so a busy supervisor (e.g. mid-`connect_once` with a hung transport factory) can't block other registry ops. The `shutdown_is_not_blocked_by_a_hung_transport_factory` test exercises this path end-to-end.

connor4312 (approved):

PR #142, opened 2026-05-21
Fix SessionReady reducer clobbering activity status mid-turn

roblourens:
## Problem

`SessionReady` is a lifecycle transition (`Creating` → `Ready`) but the reducer was also hardcoding `summary.status = SessionStatus.Idle`. For **provisional sessions**, the first turn can start before materialization completes, so by the time `SessionReady` is dispatched from the materialize-session handler, an `activeTurn` may already exist and the status may already be `InProgress`. `SessionReady` would then briefly clobber that to `Idle`, and the next tool-call lifecycle action (which calls `refreshSummaryStatus`) would flip it back to `InProgress` — producing a spinner-flicker for clients that bind directly to `summary.status`.

Observed sequence from a VS Code repro of [microsoft/vscode#316543](https://github.com/microsoft/vscode/issues/316543) with diagnostic logs:

```
21:07:36.560  handler  _handleTurn starting               (dispatches SessionTurnStarted)
21:07:36.898  list     sessionAdded   protocolStatus=InProgress(0x8)
21:07:36.999  list     sessionSummaryChanged  protocolStatus=Idle(0x1)   ← bug (~100ms later)
21:07:39.838  list     sessionSummaryChanged  protocolStatus=InProgress(0x8)  ← refreshSummaryStatus flips back
21:07:47.256  handler  _handleTurn onTurnEnded
21:07:47.356  list     sessionSummaryChanged  protocolStatus=Idle(0x1)   ← legitimate end-of-turn
```

## Fix

Every other reducer that touches activity-related data already keeps `summary.status` in sync via `summaryStatus`/`refreshSummaryStatus`. `SessionReady` therefore doesn't need to recompute the status — it just shouldn't touch it at all. For non-provisional sessions (where no turn is active when `SessionReady` runs) this is equivalent to the previous behavior, since the status will already be `Idle`.

## Test

Added fixture `147-session-ready-preserves-inprogress-status.json` that exercises the provisional-session ordering (`activeTurn` set + `status = InProgress` before `SessionReady`) and verifies the status is preserved. Verified the new fixture fails on `main` and passes with this change. Reducer coverage stays at 100% branches.

(Written by Copilot)

### DIFF
--- a/clients/rust/crates/ahp/src/reducers.rs
+++ b/clients/rust/crates/ahp/src/reducers.rs
@@ -432,8 +432,9 @@ pub fn apply_action_to_root(state: &mut RootState, action: &StateAction) -> Redu
 pub fn apply_action_to_session(state: &mut SessionState, action: &StateAction) -> ReduceOutcome {
     match action {
         StateAction::SessionReady(_) => {
+            // Lifecycle-only transition. Must not touch `summary.status`: see
+            // the equivalent TypeScript reducer for the rationale.
             state.lifecycle = SessionLifecycle::Ready;
-            state.summary.status = summary_status(state, None);
             ReduceOutcome::Applied
         }
         StateAction::SessionCreationFailed(a) => {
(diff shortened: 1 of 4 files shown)


connor4312 (approved):
ah, that was the issue

PR #155, opened 2026-05-27
Rename UserMessage → Message; add MessageKind discriminant

roblourens:
Renames `UserMessage` → `Message` and `userMessage` → `message` throughout the protocol (state, actions, commands), and introduces a `MessageKind` discriminant carried on each message.

### Changes

- **Types**: rename `UserMessage` to `Message` (state) and rename the `userMessage` field on `SessionTurnStartedAction`, `SessionPendingMessageSet`, `ActiveTurn`, `Turn`, and `PendingMessage` to `message`.
- **`MessageKind` enum** (`"user" | "systemNotification"`) and a new required `Message.origin: { kind: MessageKind }` field — messages now carry an explicit origin discriminant instead of being implicitly user-only.
- Updates reducers, all reducer fixtures under `types/test-cases/reducers/`, the JSON Schema, generators, and regenerated Swift/Rust/markdown.
- **Swift client** (hand-written): `Reducers`, `NativeReducer`, `AHPApp` store/views, and tests updated to construct `Message(text:, origin:, …)`.
- **Rust client** (hand-written): `ahp::reducers` production and test sites updated to use `Message { text, origin, attachments }`.

### Open questions / follow-ups

- `Message.origin` is currently an anonymous inline TS type, so it generates as `AnyCodable` in Swift and `AnyValue` (`serde_json::Value`) in Rust. A follow-up could name it as a proper `MessageOrigin` interface so clients get a typed shape and discriminated-union ergonomics.

(Written by Copilot)


types/channels-session/state.ts
```diff
@@ -600,18 +600,31 @@ export interface ActiveTurn {
...
  * in the text.
  *
  * @category Turn Types
  */
-export interface UserMessage {
+export interface Message {
   /** Message text */
   text: string;
+  /** The origin of the message */
+  origin: { kind: MessageKind };
```
connor4312, line 627: In the case of a SystemNotification, should we provide an alternate text to show in place of the model `text`?

roblourens: The model text wouldn't be in here, because the client doesn't dispatch this, so it would just be the display text

Yoyokrazy (approved):

PR #166, opened 2026-05-28
Add Go client to clients/go/

colbylwilliams:
## Summary

Adds a sixth shippable artifact: a **Go module** at `clients/go/`, mirroring the existing Rust / Kotlin / Swift / TypeScript client surface area. The module is published as `github.com/microsoft/agent-host-protocol/clients/go` via the standard Go module proxy.

The Go client ships:

- **`ahptypes`** — generated wire types (string + bitset enums, discriminated unions as concrete wrapper structs with custom `Marshal/UnmarshalJSON`, `Partial<T>` synthesized with pointer fields, forward-compat `XUnknown` variants).
- **`ahp`** — async `Client` over a pluggable `Transport`, four pure reducers (`ApplyActionToRoot/Session/Terminal/Changeset`), `MultiHostStateMirror` helper, centralized `Done()` / `Err()` shutdown lifecycle.
- **`ahp/hosts`** — `MultiHostClient` with reconnect supervisor, `ClientIDStore` (in-memory and file-backed), `HostHandle` snapshots, generation-checked `HostClientHandle`.
- **`ahpws`** — WebSocket transport on `github.com/coder/websocket`.
- **`examples/`** — `connect_ws` and `reducers_demo`.

The new `scripts/generate-go.ts` is wired into `scripts/generate.ts` and exposed as `npm run generate:go`.

## Why

Brings Go into the protocol-SDK family alongside the four existing clients, sharing the same single source of truth in `types/*.ts` and the same fixture-driven parity guarantee. Go consumers (e.g. agent hosts, server-side tooling, CLI runtimes) no longer need to hand-roll JSON-RPC plumbing.

## Validation

- **158 / 158** cross-language reducer fixtures under `types/test-cases/reducers/` pass via the new `TestFixtureDrivenReducerParity` test — same suite the Rust / Kotlin / Swift clients use.
- Client and multi-host smoke tests pass under `go test -race ./...`.
- `npm run generate:go` is idempotent (CI gates on drift via `git status --porcelain`).
- `npm run typecheck`, `npm run lint`, `npm run verify:release-metadata`, `npm run verify:changelog`, and the full `npm run generate` all pass.

## Design decisions

- **Tag namespace `clients/go/vX.Y.Z`** (not `go/vX.Y.Z` and not bare semver). The Go module proxy requires the tag prefix to match the module's directory inside the repo for sub-module resolution (`go help mod` → "Module versions"). Bare semver tags remain reserved for Swift.
- **Version source = `clients/go/VERSION`** (mirrors Swift). `go.mod` has no version field; the version file is what `release-metadata.json`, `verify-changelog.ts`, and `publish-go.yml` cross-check against.
- **Discriminated unions emitted as concrete wrapper structs** (`type StateAction struct { Value isStateAction }`) rather than bare marker interfaces, so a wrapper-typed field like `Turn.ResponseParts []ResponsePart` round-trips through `encoding/json` without consumer-side custom marshalers.
- **`omitempty` only on optional pointer fields** so required `serverSeq: 0` and required empty arrays don't get silently elided.
- **Reducers mutate `*State` in place** to match the Rust client's semantics and let the fixture runner compare final JSON.
- **`MultiHostClient` reconnect supervisor uses `Client.Done()`** rather than a polling ping loop, so reconnection is event-driven and shutdown is bounded.

## CI / release

- `.github/workflows/ci.yml` gains a `go` job pinned to Go 1.22 with the standard generator-drift check.
- `.github/workflows/publish-go.yml` triggers on `clients/go/vX.Y.Z` tags, re-runs every drift / metadata / changelog gate, validates tag ↔ `VERSION`, warms the Go module proxy with the new version, and creates a GitHub Release whose body is the matching CHANGELOG section.
- `RELEASING.md` and root `AGENTS.md` are updated for the sixth artifact (tag conventions, CHANGELOG map, secrets table).

## Out of scope (intentional)

- The changeset reducer is a `NoOp` stub, matching the Rust client's current deferred state.
- `HostHandle.{Agents, Sessions, Terminals}` fields are reserved for a later release that wires the per-host root reducer into the supervisor — currently always nil. Consumers should subscribe via `MultiHostClient.Subscriptions()` and feed events through `ApplyActionToRoot` in the meantime.


colbylwilliams:
unfortunately CCR keeps erroring

connor4312 (approved):

Issue #162, opened 2026-05-28
Proposal: sessionFs/* server → client request family for client-provided per-session filesystem


colbylwilliams:
> **Update 2026-05-28**: Original body proposed a new `sessionFs/*` server → client request family. @connor4312 [pointed out](https://github.com/microsoft/agent-host-protocol/issues/162#issuecomment-4564826537) that the `resource*` family is already bidirectional in practice — VS Code's `_handleReverseRequest` in `src/vs/platform/agentHost/browser/remoteAgentHostProtocolClient.ts` implements `resourceList` / `resourceRead` / `resourceWrite` as reverse RPCs, and the customizations-sync flow uses them that way to fetch `virtual://my-client/...` content from the publishing client. The types just haven't caught up. This rewrite reframes the proposal around **formalizing the existing pattern + adding the missing methods**. SQLite is carried forward as an open question.

## Summary

The Copilot SDK's [`SessionFsProvider`](https://github.com/github/copilot-sdk/blob/main/rust/src/session_fs.rs) trait lets host applications back the CLI's per-session filesystem with a custom store (sandbox, in-memory, remote, audited). To plumb that semantics through AHP, the host needs to invoke filesystem operations on whichever connected AHP client owns the workspace for a given session. AHP's `resource*` family is already used this way in practice (per Connor's note above) but:

1. The typed `ServerCommandMap` reflects only `resourceRequest`, not the rest of the family, so the bidirectional usage is undocumented and untypeable.
2. The family is missing several methods the SDK trait surfaces — `stat`, `mkdir`, `append`, `exists` — and SQLite, which doesn't fit `resource*` semantics at all.

This issue proposes: (a) formalize the de facto bidirectional pattern in `ServerCommandMap`, (b) add the missing file-shaped methods, (c) decide where SQLite belongs, (d) document the per-session "which client owns the FS" convention.

## Motivation

`copilot-sdk` ships a `SessionFsProvider` trait (PR [github/copilot-sdk#1428](https://github.com/github/copilot-sdk/pull/1428), part of `ClientMode::Empty`). When the SDK is started with `ClientOptions::session_fs` populated, it calls `sessionFs.setProvider` on the CLI during `Client::start`, telling the CLI to delegate **all per-session filesystem operations** to a host-supplied provider.

### SDK wire surface (CLI → SDK), with current AHP coverage

| SDK method                   | AHP equivalent today             | Status |
| ---------------------------- | --------------------------------- | ------ |
[...]

PR #167, opened 2026-05-28
resource*: add resolve/mkdir/watch, write modes, Conflict code

connor4312:
Closes the gaps surfaced in #162 by extending the existing `resource*` family rather than introducing a parallel namespace, and formalizing that the whole family is bidirectional (VS Code already uses it that way for client-published `virtual://` URIs).

New methods (`Client ↔ Server`):

- `resourceResolve` — stat + realpath combo. Returns canonical URI, type, size, mtime, ctime, contentType, and an opaque `etag` for optimistic concurrency. Throws `NotFound` for the "does this exist" check — no separate `resourceExists` shim.
- `resourceMkdir` — `mkdir -p` only. Idempotent on existing directories; fails with `AlreadyExists` only when `uri` exists as a non-directory.
- `createResourceWatch` — opens a file-change watcher and returns a fresh `ahp-resource-watch:/<id>` channel URI. Subscribers receive `resourceWatch/changed` actions; unsubscribing the last subscriber releases the watcher (no explicit dispose call).

New `resourceWrite` fields:

- `mode: ResourceWriteMode` (`truncate` | `append` | `insert`) plus `position?: number`. Truncate is rooted at start of file, append is rooted at EOF (position counts back from EOF), insert is rooted at start of file. Default behaviour (mode omitted) is the existing full overwrite.
- `ifMatch?: string` for optimistic concurrency against `resourceResolve.etag`. Failure surfaces as the new `Conflict` (-32011) error code.

Existing `resourceRead/Write/List/Copy/Delete/Move` methods promoted to `ServerCommandMap` and `@direction` JSDoc updated to `Client ↔ Server` — matches VS Code's `_handleReverseRequest`, which has been serving the customizations-sync flow this way since v0.1.

Wiring:

- New `types/channels-resource-watch/` channel module (state, action, reducer, commands) following the same shape as channels-terminal.
- `ActionType.ResourceWatchChanged` registered at `0.2.0`; protocol version itself unchanged per general-instructions.
- All four code generators (TypeScript/Swift/Rust/Kotlin) and the action-origin generator extended; `ResourceWriteMode` and `ResourceType` exported as const enums for cross-language parity.
- `subscriptions.md`, `overview.md`, `root-channel.md` updated; new `resource-watch-channel.md` spec page added to the sidebar.

Tests: 225 pass, 100% reducer branch coverage. Two new reducer fixtures cover the watch reducer (event passthrough + unknown action no-op).


colbylwilliams (commented):
This is great — the bidirectional `resource*` formalization, `resourceResolve` (stat + realpath in one), `ResourceWriteMode`, and the `ifMatch`/`etag` story all close real gaps cleanly.

One thing that may be a wire-shape bug, or an undocumented intentional convention — flagging either way:

**`subscribe`/`unsubscribe` aren't in `ServerCommandMap`, but the watcher flow requires server-initiated `subscribe`.**

`createResourceWatch` is symmetrical — either peer MAY initiate (`types/common/messages.ts:184`, plus the `@direction Client ↔ Server` JSDoc in `types/channels-resource-watch/commands.ts:31`). The watcher-channel spec then says:

> The caller then [`subscribe`](./subscriptions)s to that channel to receive `resourceWatch/changed` actions over the standard action envelope.  
> — `docs/specification/resource-watch-channel.md` (and the lifecycle diagram)

For the server-initiated case (host wants to observe a client-side URI — e.g. host-driven per-session FS providers, the use case the PR description calls out), the **server** is the caller and would need to `subscribe` to the receiver-allocated `ahp-resource-watch:/<id>` channel on the client. But `subscribe` lives only in `CommandMap` (`types/common/messages.ts:148`) and `unsubscribe` only in `ClientNotificationMap` (`:211`). The server-as-caller path has no wire shape to subscribe back.

Two reasonable fixes:

1. **Promote `subscribe`/`unsubscribe` to `ServerCommandMap`/`ServerNotificationMap` alongside `createResourceWatch`.** Smallest spec surface, matches the precedent set by the `resource*` family in this same PR (any command that can be server-initiated has its subscribe/unsubscribe lifecycle exposed too). Probably the cleanest answer.

2. **Document an implicit-subscribe convention for receiver-allocated channels.** Add normative text along the lines of: *"When a receiver allocates a channel URI in response to a `createResourceWatch` request from a caller, the caller is implicitly subscribed for the lifetime of the watch; no explicit `subscribe` is required, and `unsubscribe` from the original caller releases the watcher."* Avoids touching `subscribe`'s direction, at the cost of `ahp-resource-watch:/<id>` having different lifecycle rules from every other channel in AHP.

Happy to be wrong about this if the intent was always (2) and it just needs a sentence — but in the current text the asymmetry is silent.

(Other gaps from #162 — SQLite, per-session FS-provider binding convention, POSIX mode parity, and disconnect-mid-call semantics — I'll leave on the issue rather than scope-creep this PR. Sections 1 + 3 of the issue are clearly closed here.)

roblourens (dismissed):

connor4312:
> For the server-initiated case (host wants to observe a client-side URI — e.g. host-driven per-session FS providers, the use case the PR description calls out), the server is the caller and would need to subscribe to the receiver-allocated ahp-resource-watch:/<id> channel

So far we haven't had a need for an agent host to watch files on the client. Is there a case you have where that would be necessary?

dmitrivMS (approved):

PR #185, opened 2026-06-03
feat: add `_meta` provider metadata to `AgentCustomization` across clients

DonJayamanne:


connor4312:
@DonJayamanne not opposed to this in principal but I'm curious why we need this already -- is there a non-standard field we want to talk about?

DonJayamanne:
> y -- is there a non-standard field we want to talk about?

Had a chat with @aeschli and almost all of the fields are non standard or we have different interpretations
E.g. event the `prompt` isn't a standard field.
We treat `model` as an array, others & spec treats it as a single value
Tools have different meanings as well.

TLDR: Based on my discussions with martin, there aren't any, hence his suggestion was to not add any property to this spec

**Phase 1**
* So i opted to go with the simplest to unblock from the perf issue of VS Code reading the file everytime
I want to get this , as we have a perf issue with VS Code (client) reading the files everytime (& lots of AHP messages sent around)

**Phase 2**
* Figure out what are the standard properties, we must have some, I'd leave phase 2 as a debt issue to remove this `content` property

connor4312 (approved):

Issue #186, opened 2026-06-03
Reducers documented as "pure" stamp summary.modifiedAt from the wall clock (Date.now); TS client lacks the now-seam the other clients have


joshmouch:
### Summary

The docs state reducers are pure and that "the same reducer code runs on both server and client, which is what makes write-ahead possible" ([`docs/guide/actions.md` L196](https://github.com/microsoft/agent-host-protocol/blob/main/docs/guide/actions.md#L196); also [`README.md` L9](https://github.com/microsoft/agent-host-protocol/blob/main/README.md#L9)). In practice the session reducer reads the ambient wall clock to stamp `summary.modifiedAt`. In the published `@microsoft/agent-host-protocol@0.2.0` TypeScript client, this is a hardcoded inline `Date.now()` with no injection seam, so an identical `(state, action)` pair yields different output depending on when it runs.

### Evidence (`types/channels-session/reducer.ts`)

10 `Date.now()` sites, all stamping `summary.modifiedAt`: L155 (`endTurn`), L175 (`upsertInputRequest`), L298 (turnStarted), **L520 (titleChanged)**, L543 (modelChanged), L549 (agentChanged), L590 (configChanged), L694 (turn truncate), L729 (inputAnswerChanged), L749 (inputCompleted). For example:

```ts
case ActionType.SessionTitleChanged:
  return { ...state, summary: { ...state.summary, title: action.title, modifiedAt: Date.now() } };
```

The action carries no timestamp (`SessionTitleChangedAction = { type, title }`), and `sessionReducer(state, action, log?)`'s third param is a logging callback, not a clock — so there is no way to make this deterministic per-call in the TS client today.

### Why this matters — field ownership

`summary.modifiedAt` reads as **server-authoritative**: the server streams it to clients via `root/sessionSummaryChanged` ([`state-model.md` L381](https://github.com/microsoft/agent-host-protocol/blob/main/docs/guide/state-model.md#L381)). But under write-ahead reconciliation the client *replays the same action* to compute `optimisticState` ([`reconciliation.md` L11, L14](https://github.com/microsoft/agent-host-protocol/blob/main/docs/guide/reconciliation.md#L11)), so the client stamps a *different* `modifiedAt` (its local `Date.now()`) than the server stamped — a transient divergence until the server echo overwrites `confirmedState`. This is the classic "one timestamp, two writers" / domain-value-vs-data-meta problem (e.g. `serverModifiedAt` vs `clientModifiedAt`).

[...]

Issue #187, opened 2026-06-03
Remove SessionSummary from SessionState


connor4312:
There are initial debt items around here...
- SessionSummary has model/agent selection, but these can probably just live on the SessionState
- #186 
- createdAt/modifiedAt is numeric on SessionSummary but an ISO timestamp on ResourceResolveResult. I think ISO is a preferred representation.
- nit: changesets on the summary is probably too verbose/noisy

...but while we're here, the whole situation of the SessionSummary being embedded on the SessionState is bad. We have two different sets of actions to update the same thing, which has led to confusion in implementation and makes it really hard to fix #186 nicely.

SessionSummary should share some fields with SessionState, but it should be its own thing that exists solely on the root. In that case the host is just in control of updating the modifiedAt at its own discretion and get we get rid of the oddly overlapping actions.

We can also reduce how much of `changesets` we want to expose on the summary.

Fix timestamps at the same time.

joshmouch:
Adding consumer-side empirical data that's relevant to how this restructure handles `modifiedAt`, since it bears directly on #186.

I built a standards-compliant AHP **consumer** against the published `@microsoft/agent-host-protocol@0.2.0` (`AhpClient` + `WebSocketTransport`, driving the published `sessionReducer`), and hit a concrete gap worth flagging here:

**The published wire `ActionEnvelope` has no host-authoritative metadata channel.** It is exactly:

```ts
// dist/types/common/actions.d.ts
export interface ActionEnvelope {
  channel: string;
  action: StateAction;
  serverSeq: number;
  origin?: ActionOrigin;
  rejectionReason?: string;
}
```

No `meta` / `modifiedAt` / `serverTimestamp`. Combined with #186 (the reducer stamping `summary.modifiedAt = Date.now()`), this means a compliant client converges on the host's authoritative `modifiedAt` **only via the snapshot / reconnect-snapshot path** — *not* on the two legs that replay bare actions:

1. **Write-ahead optimistic apply** — the client replays its own pending actions over `confirmedState` (`docs/guide/reconciliation.md`), re-running the reducer → re-stamping `modifiedAt` from the *client's* clock. No snapshot in this leg.
2. **Reconnect-replay** — the `ReconnectResult.Replay` branch streams missed action envelopes; those carry no timestamp, so the reducer re-derives `modifiedAt` again.

So #186's now-seam makes `modifiedAt` *deterministic per client*, but nothing on the wire makes it *converge to the host's value* on those replay legs.

[...]

PR #189, opened 2026-06-03
TS client: injectable currentTimestampProvider for reducer modifiedAt (#186)

joshmouch:
## Summary

Brings the **TypeScript** client to parity with Go/Kotlin/Rust by adding an injectable now-seam for the `summary.modifiedAt` timestamps the session reducer stamps. Closes the TS half of #186.

### The problem (#186)
The docs state reducers are pure and "the same reducer code runs on both server and client, which is what makes write-ahead possible" — but `types/channels-session/reducer.ts` stamps `summary.modifiedAt = Date.now()` at 10 sites, so an identical `(state, action)` pair yields different output depending on *when* it runs. Under write-ahead reconciliation the client replays the same action and stamps a *different* `modifiedAt` than the server, until the server echo overwrites it.

The Go, Kotlin, and Rust clients already deliberately stamp from a clock **but expose an injectable seam** for deterministic replay/tests (Go `SetNowProvider`, Kotlin `currentTimestampProvider`, Rust `MOCK_NOW_MS`). The TS client was the only one with **no seam**, forcing its own fixture test to monkeypatch the process-global `Date.now` (racy under concurrency, not per-call controllable).

### The change (non-breaking)
- `types/channels-session/reducer.ts`: add `currentTimestampProvider: () => number` (default `Date.now`) + `setCurrentTimestampProvider(...)`, mirroring Kotlin's `currentTimestampProvider`. Replace the 10 inline `Date.now()` sites with `currentTimestampProvider()`.
- Re-exported through `types/reducers.ts` and `types/index.ts`.
- `types/reducers.test.ts`: inject the seam (`setCurrentTimestampProvider(() => 9999)`) in `beforeEach` instead of monkeypatching the global `Date.now` — matching the other clients' fixture runners.

Default behavior is unchanged (`Date.now`), so this is purely additive for existing consumers.

### Verification
- `tsc --noEmit -p types/tsconfig.json` — clean.
- `tsx --test types/reducers.test.ts` — **169 pass / 0 fail** (incl. the 163 shared reducer fixtures, now driven through the seam).

### Note on intent
Filed issue-first as #186, which asked whether you'd prefer (a) docs clarification, (b) TS now-seam parity, or (c) a wire-breaking timestamp change. This PR implements **(b)** — the non-breaking option that matches the existing cross-client precedent — but happy to adjust toward (a)/(c) or a different seam name/shape per your preference.


joshmouch:
@microsoft-github-policy-service agree

joshmouch:
Thanks @connor4312 — that makes sense, and I agree #187 is the better path. Moving `modifiedAt` to a host-authoritative spot on the root summary removes the need for a client-side now-seam altogether (and it'd actually close the optimistic-vs-confirmed convergence gap I flagged over on #187, not just make the client's re-derivation deterministic). So I'll close this in favor of that.

If a short docs clarification would be useful in the meantime, I'm happy to send a small one — otherwise I'll leave the timestamp work to #187. Thanks for taking a look!

PR #190, opened 2026-06-03
docs: add Go client to the root README clients list

joshmouch:
The Go client (`clients/go`) is on `main` with its own `README`, `CHANGELOG`, `AGENTS.md`, and a `publish-go.yml` workflow, but the root `README.md` "Clients" list — and the multi-host paragraph — were never updated to mention it. Every other first-party client (Swift, Rust, Kotlin, TypeScript) is listed; Go is the only one missing.

This PR adds:
- A **Go** bullet to the Clients list, matching the style of the others: `go get` line, the three packages (`ahptypes` / `ahp` / `ahpws`), a link to `clients/go/` + its CHANGELOG, and a note that releases happen via `clients/go/vX.Y.Z` tags through the Go module proxy (no separate registry).
- Go to the multi-host paragraph (the Go SDK ships `MultiHostClient` in `ahp/hosts`, with `hosts.Single(...)` for the single-host case).

Docs-only; no code or generated output changes, so no CHANGELOG entry per `AGENTS.md`.

> Note for maintainers: there is no `clients/go/v*` tag yet (current tags are `rust/*`, `spec/*`, `typescript/*`), so the first Go release tag still needs to be pushed to trigger `publish-go.yml`. This README entry documents the consumption path that is already wired up.


connor4312 (approved):
thanks!

rzhao271 (approved):

PR #201, opened 2026-06-08
Don't run generator when formatter isn't available

benibenj:
This has cause a lot of pain for me. I did not notice that I didn't have the formatters installed and so I ended up having CI builds red all the time without understanding why. As the generated files must be formatted before being able to merge them I would suggest throwing when none is installed and providing a command line argument instead to be very clear that something is not setup correctly.
I also added some other default paths on windows for go formatter

### DIFF
--- a/.github/workflows/format-generated.yml
+++ b/.github/workflows/format-generated.yml
@@ -1,0 +1,77 @@
+name: Format Generated Sources
+
+# Manually-triggered workflow for maintainers who don't have the Rust
+# (`cargo fmt`) or Go (`gofmt`) toolchains installed locally and therefore
+# generated the clients with `--allow-missing-formatter` (leaving the
+# generated Rust/Go sources unformatted).
+#
+# It checks out the branch it was dispatched against, installs the Rust and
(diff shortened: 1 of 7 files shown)


connor4312:
I wonder if we can also have a CI action that developers can manually run that formats + commits to their branch, similar to what we had in copilot-chat with the update simulation cache pipeline. And we could reference that as an alternative in the output warning

benibenj:
That would be neat. But are you referring to maintainers or contributors? For maintainers that should be straight forward to add. But if you are referring to contributors without write permissions, I'm not sure how they would be able to trigger this.

connor4312:
Yea, maintainers only, sorry

benibenj:
I added a github action to format `rust` and `go` client code. I tested it and it worked, you can check it here: [cb5568f](https://github.com/microsoft/agent-host-protocol/pull/201/commits/cb5568f60fa4c9c0a16e8c5fe3e4525c838d581e)

In theory we could just run it automatically, but I guess there might be scenarios where this would be annoying.

connor4312 (approved):
nice!

PR #204, opened 2026-06-09
Add a shared round-trip corpus + fix the wire-fidelity bugs it caught (SessionStatus bitset, changeset range, origin)

joshmouch:
## What this does

Adds a shared, language-agnostic **round-trip corpus** under `types/test-cases/round-trips/` that every client (Go, Rust, Swift, Kotlin, TypeScript) decodes into its real generated types, re-encodes, and asserts against a single canonical form — the wire-fidelity analog of the existing reducer fixtures.

Standing it up surfaced three real bugs where the generated clients had drifted from the canonical TS source, so this PR also fixes them.

### ⚠️ Two SemVer-major fixes (the clients were already non-conformant)

- **SessionStatus** is a numeric bitset on the wire (the `Bitset` JSDoc marker in `types/channels-session/state.ts`), but Rust modeled it as a closed `enum` and Kotlin as a signed `Int` — neither could hold a flag combination (`InProgress | IsArchived` = 72) or an unknown forward-compat bit (e.g. `2147483720`). It is now a uniform **32-bit unsigned** bitset: Rust `u32`, Kotlin `UInt`, Swift `UInt32` (Go was already `uint32`).
- **`ChangesetOperationTarget` range** is `range: TextRange` per the spec (nested `{start: {line, character}, end: {line, character}}`), but the generators hardcoded a flat `{start, end}` integer pair that cannot represent a real range. It now references the canonical `TextRange` (TypeScript already did).

### Non-breaking fixes

- `ActionEnvelope.origin` is omitted (not `null`) when absent, per the spec (`origin: ActionOrigin | undefined`; `undefined` = server-originated).
- Forward-compat: unknown discriminants on evolvable unions round-trip instead of throwing.

## Corpus conventions

- Each fixture is one `input` object + an `acceptableOutputs` array (per @connor4312's suggestion on the previous review). Each array carries **exactly one** canonical entry — multiple "acceptable" outputs would risk cementing observed-but-wrong client divergence instead of pinning the single intended wire form.
- **Exact-match**, key-order-independent but **key-presence-sensitive**: `null` is not normalized to absent, so e.g. the `origin` omit-vs-`null` distinction is genuinely tested.
- **Group A** = every client agrees. **Group B** (017/019) = a known type carries unknown wire keys; runtime decoders drop them (`acceptableOutputs[0]`) and TypeScript preserves them (`typescriptOutput`) — a real type-system difference, documented in `KNOWN-FIDELITY-GAPS.md`, never a skipped assertion.

## On scope

The corpus is additive; the two SemVer-major fixes are coupled to it (the corpus is what caught them, and their regression fixtures cannot pass without the fixes). I have kept them together as one coherent change, but I am very happy to split the breaking fixes into separate PRs if you would rather review or version them independently — just say the word.


types/test-cases/round-trips/019-channel-scoped-notification-uri.json
```diff
@@ -0,0 +1,27 @@
...
+    "unknownFutureKey": { "nested": true }
+  },
+  "expect": {
+    "channel": "ahp:/root",
+    "summary.resource": "ahp-session:/s1",
+    "summary.provider": "demo",
+    "summary.title": "New session",
+    "summary.createdAt": 1,
+    "summary.modifiedAt": 2
+  },
```
connor4312: I'm not a huge fan of the varied types of expectations here. I think these just make the tests a bit harder for an LM to author and test against consistently in other languages. It's a bit more verbose, but maybe the tests at modelled with one `input` object and then `acceptableOutputs: object[]`

joshmouch:
Done. Every fixture is now a single `input` + `acceptableOutputs` array, uniform across the five clients.

I held `acceptableOutputs` to one entry on purpose: allowing several "acceptable" forms lets clients diverge instead of pinning a single wire shape, which is the whole point of the corpus. The one case where clients legitimately differ (a known type carrying unknown future keys, which runtime decoders drop and TS can't) is a `group: "B"` + `typescriptOutput` field, so that type-system difference is explicit rather than a second blessed output.

types/test-cases/round-trips/KNOWN-FIDELITY-GAPS.md
```diff
@@ -0,0 +1,48 @@
...
+## Group A vs Group B
+
+- **Group A** (`"group": "A"`, or absent): every client agrees; all assert
+  `acceptableOutputs[0]`.
+- **Group B** (`"group": "B"`): a known type carries extra, unmodeled wire keys.
+  Runtime-decoder clients (Go, Rust, Swift, Kotlin) decode into a typed struct,
+  which drops the unknown keys, and assert the dropped form in
+  `acceptableOutputs[0]`. TypeScript has no runtime decoder, so `JSON.parse` /
+  `JSON.stringify` preserve every key; it asserts the preserved form in
+  `typescriptOutput`. TypeScript still asserts — it is never skipped. Fixtures
```
connor4312: should typescriptOutput jsut be one of the acceptableOutputs?

joshmouch: I tried that first. Folding it into `acceptableOutputs: [dropped, preserved]` makes the test weaker: each client passes if it matches either form, so it can't catch a runtime client that drifts to preserving unknown keys, or the TS path drifting to dropping. The split pins each client to the one form it should produce, so a change to that behavior is a deliberate corpus edit instead of a silent pass.

That's worth pinning because preserve-vs-drop is observable, not cosmetic. Say v0.4 adds `pinnedAt` to `SessionSummary` and a v0.3 client receives it: the runtime clients decode into a struct and drop `pinnedAt` on re-encode, while the TS client (no runtime schema) keeps it. Anywhere decoded state gets re-serialized (persisted and reloaded, crossing a worker/SSR boundary, relayed onward), `pinnedAt` survives the TS path and not the others, and over multiple hops it survives only if every hop preserves. A consumer can see different data depending on which client was in the path.

One thing worth flagging: the spec doesn't actually settle this either way. Versioning only says to ignore unknown action *types*, not how to handle unknown *keys* on a known type. So if you'd rather all clients converge (all drop, or all preserve for forward-compat passthrough), I'm glad to make that the rule and assert a single form everywhere. Until then, the split keeps the divergence visible and tested rather than unspecified.

connor4312: Let's name this at least `preservedOutput` since other clients might eventually use it

connor4312 (approved):
thanks!

types/test-cases/round-trips/KNOWN-FIDELITY-GAPS.md
joshmouch: That's a much better name, thanks!

TylerLeonhardt (approved):

PR #205, opened 2026-06-09
Implement Swift changeset + resourceWatch reducers; un-skip terminal fixtures

joshmouch:
The Swift client shipped only 3 of the 5 state reducers (root, session,
terminal). changesetReducer and resourceWatchReducer were missing, and the
fixture-driven parity test silently skipped the terminal, changeset, and
resourceWatch fixtures with a bare continue — even though terminalReducer
already existed. That left the shared cross-language fixtures for those three
families unverified against Swift.

- Port changesetReducer mirroring the canonical TypeScript reducer (and the
  Kotlin/.NET clients): statusChanged, fileSet, fileRemoved, operationsChanged,
  operationStatusChanged, cleared. Carry error only on the error status so a
  recovered changeset/operation does not keep a stale error.
- Port resourceWatchReducer as the documented event-pass-through (changed
  leaves state unchanged; unknown actions are a no-op).
- Remove the blanket skip and wire the terminal/changeset/resourceWatch arms so
  all three families decode and assert. Replace the skip with a drift tripwire
  (an explicit known-gap set asserted to match exactly), mirroring the
  round-trip corpus test, plus a guard pinning the previously-skipped count.

Two documented gaps remain in the known-gap set:

- Representational: fixture 103 feeds a response part with an unknown kind,
  which the generated ResponsePart enum on this base rejects. It closes
  automatically once the generated types gain an unknown-discriminant fallback.
- Unimplemented channel: the annotations channel (fixtures 210–217) has no
  Swift reducer yet, so it lands on the runner's default arm. The prior test
  skipped the annotations family with a bare continue; here that skip is made
  explicit and tripwired instead. It closes once annotationsReducer lands.

swift test passes; the runner now exercises the terminal, changeset, and
resourceWatch fixtures (32 cases) that were previously skipped.

### DIFF
--- a/clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/Reducers.swift
+++ b/clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/Reducers.swift
@@ -948,3 +948,90 @@ public func terminalReducer(state: TerminalState, action: StateAction) -> Termin
         return state
     }
 }
+
+// MARK: - Changeset Reducer
+
+/// Pure reducer for changeset state. Handles all changeset-scoped actions.
+///
+/// Per the spec, every changeset action is server-only. New files are
+/// appended via `changeset/fileSet` when the id is unknown and replaced in
+/// place when it matches an existing entry, preserving a stable file order.
+public func changesetReducer(state: ChangesetState, action: StateAction) -> ChangesetState {
+    switch action {
+    case .changesetStatusChanged(let a):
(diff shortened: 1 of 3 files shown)


connor4312 (approved):
thanks!

joshmouch:
> thanks!

No problem! :)

connor4312 (dismissed):

Yoyokrazy (approved):

PR #208, opened 2026-06-09
[Superseded by #204] Widen SessionStatus to a u32 bitset (Rust + Kotlin) — SemVer-major

joshmouch:
> 🔁 **Superseded by #204** (closed 2026-06-10 — not abandoned). #204 carries the same `SessionStatus` u32 bitset widening (Rust + Kotlin) plus the shared round-trip corpus, and is current with `main`. This branch was opened on a pre-#211 base, so against current `main` it would revert #211's reducer `_meta` propagation; the bitset work has been consolidated into #204 instead.

---

## Widen `SessionStatus` to a u32 bitset (Rust + Kotlin) — SemVer-major

`SessionStatus` is an unsigned 32-bit **wire bitset**: named status flags combine bitwise, and a newer host can set forward-compat bits a client version doesn't model yet. To round-trip those bits losslessly, the in-memory type must be able to hold any `u32`. The previous Rust `#[repr(u32)] enum` (a closed set) and Kotlin `Int` (signed — overflows at bit 31) couldn't, so unknown/combined bits were silently dropped on decode→re-encode.

This widens both as **one logical change**:

- **Rust:** `SessionStatus` becomes `struct SessionStatus(pub u32)` with named flag constants — combine with `|`, test membership with `.contains(..)`, read the raw value (including unknown forward-compat bits) with `.bits()`.
- **Kotlin:** `SessionStatus.rawValue` becomes `Long` (was `Int`); flag constants are `Long` literals; serialized as a JSON number.

Both are **SemVer-major**.

> **Review note:** the `types/test-cases/round-trips/` fixtures in this diff are **shared with #204** (byte-identical) and deduplicate on merge — they're best reviewed there. **This PR's unique change** is the widening itself: `clients/rust/`, `clients/kotlin/`, `scripts/generate-{rust,kotlin}.ts`. The two PRs are independent and merge cleanly in either order; this one closes the Rust/Kotlin round-trip gaps against fixtures 004/005/016.

### Verification

| client | result |
|---|---|
| Rust | `cargo test` 72 passed · `cargo fmt --check` clean · `cargo clippy` clean |
| Kotlin | `gradlew test --rerun-tasks` — 254/254 |


joshmouch:
Closing in favor of #204.

This branch was opened before #211 (`reducers: propagate tool call action metadata`) landed on `main`, so when diffed against current `main` it reverts #211's `_meta` propagation across TypeScript (`types/channels-session/reducer.ts` — removes the `tcBaseWithMeta` helper), Go (`clients/go/ahp/reducers.go`), and Swift (`Reducers.swift`), and removes the associated reducer fixtures (deletes `220-toolcall-actions-update-meta.json`; strips `_meta` expectations from `026`/`070`).

#204 carries the same `SessionStatus` u32 bitset widening (Rust + Kotlin) and the shared round-trip corpus, is current with `main` (includes #211), and leaves the reducer `_meta` work intact. Consolidating the bitset + corpus work there.

PR #210, opened 2026-06-09
swift: DRY the fixture-driven reducer test + fix two stale comments

joshmouch:
Follow-up to #205. Three cleanup fixes to `clients/swift/AgentHostProtocol/Tests/AgentHostProtocolTests/FixtureDrivenReducerTests.swift`:

1. **DRY with a generic `compareFixture` helper** — the encode→decode→fold-actions→assert-equal-JSON block was repeated verbatim for each of the five implemented reducer arms (root / session / terminal / changeset / resourceWatch). Extracted into `compareFixture<S: Codable>(file:fixture:stateType:reduce:)`, mirroring the Kotlin `FixtureDrivenReducerTest.compareFixture` pattern. Each per-variant arm is now a one-liner. Behavior is identical: same fixtures loaded, same assertions, same known-gap handling.

2. **Fix stale gap-range comment** — the annotations known-gaps comment cited fixtures `210–217`; the gap set actually includes 218 and 219 as well. Corrected to `210–219`.

3. **Replace absent-file citation** — a comment referenced `TypesRoundTripFixtureTests` as the precedent for the drift-tripwire discipline, but that file is not in this repo. Replaced with a description of the discipline itself (explicit known-gap set + exact-match drift tripwire).

No behavior change; `swift test` green (96 tests, 0 failures).

### DIFF
--- a/clients/swift/AgentHostProtocol/Tests/AgentHostProtocolTests/FixtureDrivenReducerTests.swift
+++ b/clients/swift/AgentHostProtocol/Tests/AgentHostProtocolTests/FixtureDrivenReducerTests.swift
@@ -86,12 +86,12 @@ final class FixtureDrivenReducerTests: XCTestCase {
 
     // MARK: - Known reducer gaps (documented, not silent)
     //
-    // Mirrors the drift-tripwire discipline in TypesRoundTripFixtureTests:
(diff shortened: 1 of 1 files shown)


connor4312 (approved):
thanks!

dmitrivMS (approved):

PR #215, opened 2026-06-11
fix(swift): correct AnyCodable NSNumber Int and Double encode fidelity

### LINKED ISSUE #123: Swift `AnyCodable.encode` corrupts `Int` to `Bool` (and `Double` to `Int`) for `NSNumber`-backed values
## Summary

`AnyCodable.encode(to:)` corrupts `Int` values to `Bool` (and `Double` to `Int`) when the wrapped `Any` is an `NSNumber`. This breaks any client pipeline that decodes wire JSON into `[String: Any]` (via `JSONSerialization`), wraps it in `AnyCodable`, then re-encodes through `JSONEncoder` — for example, the receive path in [`AgentHostProtocolClient`](https://github.com/microsoft/agent-host-protocol/pull/122) had to bypass `AnyCodable` entirely and route raw `Data` slices through `JSONSerialization` to avoid this. The fix is local to `AnyCodable.swift`.

The decode side is unaffected: `JSONDecoder` is strict and does not promiscuously bridge `1`/`true` to the wrong type.

## Repro

```swift
import Foundation

// AgentHostProtocol/AnyCodable.swift, verbatim — assume `AnyCodable` is in scope.

// Simulate the receive pipeline: a typed value is encoded by the peer to
// JSON, parsed back via JSONSerialization (which produces NSNumber for
// numerics), wrapped in `AnyCodable`, and re-encoded.
struct Payload: Codable { let serverSeq: Int }
let original = Payload(serverSeq: 1)

let wireBytes = try JSONEncoder().encode(original)                         // {"serverSeq":1}
let object = try JSONSerialization.jsonObject(with: wireBytes)             // [String: Any] with NSNumber
let wrapped = AnyCodable(object)
let reEncoded = try JSONEncoder().encode(wrapped)
print(String(data: reEncoded, encoding: .utf8)!)                           // ⇒ {"serverSeq":true}   ❌

// Decoding the corrupted bytes back into the typed shape fails:
do {
    _ = try JSONDecoder().decode(Payload.self, from: reEncoded)
} catch {
    print(error)                                                           // typeMismatch: Expected Int but found bool
}
```

Output:
```
{"serverSeq":true}
typeMismatch(... debugDescription: "Expected to decode Int but found bool instead." ...)
```

A `Double`-valued `NSNumber` is also corrupted to `Int` for the same reason (`as Int` matches before `as Double`).

## Diagnosis

```swift
public func encode(to encoder: Encoder) throws {
    var container = encoder.singleValueContainer()
    switch value {
    case is NSNull:
        try container.encodeNil()
    case let bool as Bool:               // ← matches FIRST for any NSNumber
        try container.encode(bool)
    case let int as Int:
        try container.encode(int)
    ...
```

`NSNumber` bridges to `Bool`, `Int`, and `Double` regardless of the underlying value:

```
NSNumber(value: 1)    as? Bool = Optional(true)   // wrong
NSNumber(value: 1)    as? Int  = Optional(1)
NSNumber(value: true) as? Bool = Optional(true)
NSNumber(value: true) as? Int  = Optional(1)      // would also be wrong if Bool case were second
```

Reordering the `case`s doesn't fix it — `case let int as Int` would then catch `NSNumber(value: true)` and encode `true` as `1`.

The reliable distinguisher is `NSNumber.objCType`:

| Underlying value | objCType |
|---|---|
| Bool (`true`/`false`) | `c` (also `B` on some platforms) |
| Int / Int32 / Int64 | `q`, `l`, `i`, `s` (signed) or `Q`, `L`, `I`, `S` (unsigned) |
| Float | `f` |
| Double | `d` |

## Suggested fix

Special-case `NSNumber` *before* the `as Bool` / `as Int` / `as Double` pattern matches:

```swift
public func encode(to encoder: Encoder) throws {
    var container = encoder.singleValueContainer()

    // NSNumber bridges promiscuously to Bool/Int/Double — pattern matching
    // alone can't distinguish a Bool-backed NSNumber from an Int-backed one.
    // Inspect objCType to dispatch faithfully to the underlying type.
    if let number = value as? NSNumber, type(of: value) != Bool.self {
        let objCType = number.objCType[0]
        switch objCType {
        case 0x63 /* 'c' */, 0x42 /* 'B' */:
            try container.encode(number.boolValue)
            return
        case 0x66 /* 'f' */, 0x64 /* 'd' */:
            try container.encode(number.doubleValue)
            return
        default:
            try container.encode(number.int64Value)
            return
        }
    }

    switch value {
    case is NSNull:
        try container.encodeNil()
    case let bool as Bool:
        try container.encode(bool)
    case let int as Int:
        try container.encode(int)
    case let double as Double:
        try container.encode(double)
    case let string as String:
        try container.encode(string)
    case let array as [Any]:
        try container.encode(array.map { AnyCodable($0) })
    case let dict as [String: Any]:
        try container.encode(dict.mapValues { AnyCodable($0) })
    default:
        throw EncodingError.invalidValue(
            value,
            EncodingError.Context(
                codingPath: encoder.codingPath,
                debugDescription: "AnyCodable cannot encode value of type \(type(of: value))"
            )
        )
    }
}
```

The `type(of: value) != Bool.self` guard preserves the existing path for native Swift `Bool` values (which aren't `NSNumber`s).

## Suggested tests

Add to `clients/swift/AgentHostProtocol/Tests/AgentHostProtocolTests/`:

```swift
func testAnyCodableEncodePreservesIntFromNSNumber() throws {
    let object = try JSONSerialization.jsonObject(
        with: #"{"x":1}"#.data(using: .utf8)!
    )
    let wrapped = AnyCodable(object)
    let bytes = try JSONEncoder().encode(wrapped)
    XCTAssertEqual(String(data: bytes, encoding: .utf8), #"{"x":1}"#)
}

func testAnyCodableEncodePreservesBoolFromNSNumber() throws {
    let object = try JSONSerialization.jsonObject(
        with: #"{"x":true}"#.data(using: .utf8)!
    )
    let wrapped = AnyCodable(object)
    let bytes = try JSONEncoder().encode(wrapped)
    XCTAssertEqual(String(data: bytes, encoding: .utf8), #"{"x":true}"#)
}

func testAnyCodableEncodePreservesDoubleFromNSNumber() throws {
    let object = try JSONSerialization.jsonObject(
        with: #"{"x":1.5}"#.data(using: .utf8)!
    )
    let wrapped = AnyCodable(object)
    let bytes = try JSONEncoder().encode(wrapped)
    XCTAssertEqual(String(data: bytes, encoding: .utf8), #"{"x":1.5}"#)
}
```

## Impact

Any consumer that goes through the `JSONSerialization → AnyCodable → JSONEncoder` path is affected. In [PR #122 (`AgentHostProtocolClient`)](https://github.com/microsoft/agent-host-protocol/pull/122), the inbound and outbound JSON-RPC paths had to be restructured to bypass `AnyCodable` and instead carry raw `Data` slices through `JSONSerialization` round-trips (which preserves `NSNumber.objCType` faithfully). Once this issue is fixed, that workaround can be replaced with the simpler `AnyCodable`-based pipeline.

The upcoming Swift `MultiHostClient` (PR 3) will also use the same workaround until this is fixed.

## Scope

This is a one-file fix in `clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/AnyCodable.swift`. No protocol or wire changes; no codegen impact. Adding the suggested tests is straightforward.

joshmouch:
Fixes #123.

## The bug

When wire JSON is parsed with `JSONSerialization` (which boxes every number and boolean as `NSNumber`) and the result is wrapped in `AnyCodable`, re-encoding through `JSONEncoder` corrupts the values: `Int` writes as `true`/`false` and `Double` writes as an integer. The `case let bool as Bool` and `case let int as Int` arms both match any `NSNumber`, and `Bool` is matched first.

## The fix

An `NSNumber` guard at the top of `encode(to:)` inspects `objCType` and dispatches to the faithful primitive (`boolValue` for `c`/`B`, `doubleValue` for `f`/`d`, `int64Value` otherwise) before the existing switch. The `type(of:) != Bool.self` check keeps native Swift `Bool` on the existing path. This is the dispatch from the issue's "Suggested fix".

## Two things worth flagging for review

- **`scripts/generate-swift.ts` is also edited, and it produces no change in the checked-in Swift.** `AnyCodable.swift` is a generator scaffold (`generate-swift.ts` writes it only if missing), so the same fix is applied to `anyCodableContent()`. That template had also drifted from the on-disk file (it was missing the `@unchecked Sendable` rationale and the `[Any]`/`[String: Any]` equality arms); those are reconciled here so a from-scratch scaffold reproduces `AnyCodable.swift` byte-for-byte. The generator diff is larger than the generated-output diff only because of that drift reconciliation.
- **The `AHPClient.swift` `JSONSerialization` bypass is intentionally left in place.** That path solves the inbound/decode side (numbers boxed before they reach `AnyCodable`), which an encode-side fix cannot reach, so the two are complementary layers, not duplicates. Removing the bypass is a separate, riskier change and is out of scope here.

## Tests

`AnyCodableTests` covers `Int`, `Bool`, and `Double` round-tripped through `JSONSerialization`, plus a native Swift `Bool` and a `Float`-backed `NSNumber`. `swift test`: 104 passed, 0 failed.

### DIFF
--- a/clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/AnyCodable.swift
+++ b/clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/AnyCodable.swift
@@ -41,6 +41,27 @@ public struct AnyCodable: Codable, @unchecked Sendable, Equatable {
 
     public func encode(to encoder: Encoder) throws {
         var container = encoder.singleValueContainer()
+
+        // NSNumber bridges promiscuously to Bool/Int/Double — pattern matching
+        // alone can't distinguish a Bool-backed NSNumber from an Int-backed one.
+        // Inspect objCType to dispatch faithfully to the underlying type.
+        // ('c' is also Int8's encoding, but JSONSerialization only ever produces
+        // 'c' for a Bool, so the JSON-decode path this type serves is unambiguous.)
+        if let number = value as? NSNumber, type(of: value) != Bool.self {
+            let objCType = number.objCType[0]
+            switch objCType {
+            case 0x63 /* 'c' */, 0x42 /* 'B' */:
+                try container.encode(number.boolValue)
+                return
+            case 0x66 /* 'f' */, 0x64 /* 'd' */:
+                try container.encode(number.doubleValue)
+                return
+            default:
+                try container.encode(number.int64Value)
+                return
+            }
+        }
+
         switch value {
         case is NSNull:
             try container.encodeNil()
--- a/clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/AnyCodable.swift
+++ b/clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/AnyCodable.swift
@@ -47,6 +47,9 @@ public struct AnyCodable: Codable, @unchecked Sendable, Equatable {
         // Inspect objCType to dispatch faithfully to the underlying type.
         // ('c' is also Int8's encoding, but JSONSerialization only ever produces
         // 'c' for a Bool, so the JSON-decode path this type serves is unambiguous.)
+        // Unsigned integer types ('C'/'I'/'S'/'L'/'Q') encode via uint64Value: a
+        // JSON integer above Int64.max is boxed as an unsigned 'Q' NSNumber, and
+        // int64Value would silently corrupt it (it does not round-trip).
         if let number = value as? NSNumber, type(of: value) != Bool.self {
             let objCType = number.objCType[0]
             switch objCType {
@@ -55,6 +58,9 @@ public struct AnyCodable: Codable, @unchecked Sendable, Equatable {
                 return
             case 0x66 /* 'f' */, 0x64 /* 'd' */:
                 try container.encode(number.doubleValue)
+                return
+            case 0x43 /* 'C' */, 0x49 /* 'I' */, 0x53 /* 'S' */, 0x4C /* 'L' */, 0x51 /* 'Q' */:
+                try container.encode(number.uint64Value)
                 return
             default:
                 try container.encode(number.int64Value)
--- a/clients/swift/AgentHostProtocol/Tests/AgentHostProtocolTests/AnyCodableTests.swift
+++ b/clients/swift/AgentHostProtocol/Tests/AgentHostProtocolTests/AnyCodableTests.swift
@@ -1,0 +1,57 @@
+// AnyCodableTests.swift — Tests for AnyCodable encode/decode correctness.
+//
+// Focused on the NSNumber-bridging bug: when a [String: Any] dictionary is
+// produced by JSONSerialization (which boxes numeric/bool values as NSNumber),
+// AnyCodable must re-encode each value to its original JSON type rather than
+// the first matching Swift pattern-match arm.
+
+import XCTest
+@testable import AgentHostProtocol
+
+final class AnyCodableTests: XCTestCase {
+
+    func testAnyCodableEncodePreservesIntFromNSNumber() throws {
+        let object = try JSONSerialization.jsonObject(
+            with: #"{"x":1}"#.data(using: .utf8)!
+        )
+        let wrapped = AnyCodable(object)
+        let bytes = try JSONEncoder().encode(wrapped)
+        XCTAssertEqual(String(data: bytes, encoding: .utf8), #"{"x":1}"#)
+    }
+
+    func testAnyCodableEncodePreservesBoolFromNSNumber() throws {
+        let object = try JSONSerialization.jsonObject(
+            with: #"{"x":true}"#.data(using: .utf8)!
+        )
+        let wrapped = AnyCodable(object)
+        let bytes = try JSONEncoder().encode(wrapped)
+        XCTAssertEqual(String(data: bytes, encoding: .utf8), #"{"x":true}"#)
+    }
+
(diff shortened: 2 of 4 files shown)


colbylwilliams:
@joshmouch thanks for this, and a heads-up on what happened here. A parallel PR (#227) fixing the same bug (#123) landed first and merged, so the core `NSNumber` `objCType` dispatch is now on `main`.

Your PR had two things #227 didn't: handling unsigned integers above `Int64.max` via `uint64Value`, and — nice catch — also patching the `anyCodableContent()` template in `scripts/generate-swift.ts` so a from-scratch regeneration doesn't reintroduce the bug. I've opened #238 to port both of those improvements on top of what merged, with an acid test (delete + regenerate + assert a clean diff) to keep the template from drifting again.

With the core fix plus your two improvements now merged/in-flight, this PR is effectively superseded — I think it can be closed, but I'll leave that call to you and the maintainers. Really appreciate the thorough approach, especially flagging the generator drift. 🙏

joshmouch:
Ha, beaten to the punch by #227. I'll take "scooped, with credit" any day. 😄 #238 looks great. The `uint64Value` arm plus re-syncing the generator template (and that delete+regen+clean-diff acid test is a really nice touch for keeping it from drifting back) cover everything I was after. Closing this out as superseded by #227 + #238. Thanks for the speedy follow-up and the kind words! 🙏

PR #244, opened 2026-06-19
Changeset - add contentChanged action

lszomoru:


types/channels-changeset/actions.ts
```diff
@@ -57,6 +57,26 @@ export interface ChangesetFileRemovedAction {
...
+ *
+ * @category Changeset Actions
+ * @version 4
+ */
+export interface ChangesetContentChangedAction {
+  type: ActionType.ChangesetContentChanged;
+  /** Full replacement file list. */
+  files: ChangesetFile[];
+  /** Full replacement operation list. Pass `undefined` to clear all operations. */
+  operations: ChangesetOperation[] | undefined;
```
connor4312: The "passing undefined" comment here should be changed -- undefined does not serialize in JSON, so it is not a 'thing' in the protocol

types/channels-changeset/actions.ts
```diff
@@ -57,6 +57,26 @@ export interface ChangesetFileRemovedAction {
...
+}
```
connor4312, line 80: I suggest we also add the equivalent property for `error?` here so it can be represented as well

types/channels-changeset/reducer.ts
```diff
@@ -51,6 +51,13 @@ export function changesetReducer(state: ChangesetState, action: ChangesetAction,
       return { ...state, files: next };
     }
 
+    case ActionType.ChangesetContentChanged: {
+      if (action.operations === undefined) {
+        return { ...state, files: action.files };
+      }
+      return { ...state, files: action.files, operations: action.operations };
+    }
```
connor4312, line 63: this reducer doesn't update `error`, please have your agent make a test case for it and update the code :)

connor4312 (approved):

PR #247, opened 2026-06-19
Add `Agent` and `Tool` `MessageKind` for non-user-initiated turns

### LINKED ISSUE #241: Add a `MessageKind` for agent/tool-initiated turns (non-user-initiated)
A turn can be initiated by something other than the user or a system notification — for example, a tool that spawns a worker chat whose first turn carries a seed prompt.

`MessageKind` currently has only `User` and `SystemNotification`, so a host has no faithful `origin.kind` for such a turn and must misrepresent it as one of the two existing kinds.

**Request:** consider an additional kind (e.g. `Agent` or `Tool`) for turns initiated by the agent/tooling rather than by the user.

colbylwilliams:
Closes #241.

## What

`MessageKind` only had `User` and `SystemNotification`, so a host had no faithful `origin.kind` for a turn initiated by the agent or its tooling — for example, a tool that spawns a worker chat whose first turn carries a seed prompt. Such turns had to be misrepresented as `User` or `SystemNotification`.

This adds two new kinds to the canonical enum in `types/`:

- `MessageKind.Agent` (`'agent'`) — initiated by the agent itself.
- `MessageKind.Tool` (`'tool'`) — initiated by a tool (e.g. seeding the first turn of a worker chat it spawned).

## Changes

- `types/channels-chat/state.ts`: add the two enum values + doc comments; update the `Message` doc comment.
- Regenerated all client mirrors and JSON schemas via `npm run generate`. Kotlin and the TypeScript client surface the new typed enum values; Rust/Go/Swift model `Message.origin` as an untyped object today, so only their doc comments ripple.
- `docs/guide/state-model.md`: document what each `origin.kind` value means.
- CHANGELOG entries under `[Unreleased]` for the spec and all five clients (a `types/**` change ripples to every client).

## Notes

No reducer changes: reducers don't branch on `MessageKind`, so the addition is purely additive.

## Validation

`npm run generate` and `npm run test` both clean (typecheck, lint, `verify:release-metadata`, `verify:changelog` pass; all 265 unit tests pass). The only `npm run test` hiccup is a pre-existing `c8`/`yargs` ESM incompatibility under Node 26 in the coverage wrapper, unrelated to this change — the tests pass when run directly.


types/channels-chat/state.ts
```diff
@@ -512,13 +512,25 @@ export interface ActiveTurn {
...
+   * Initiated by the agent itself rather than the user — for example, an
+   * agent that seeds the first turn of a chat it spawned.
+   */
+  Agent = 'agent',
+  /**
+   * Initiated by a tool rather than the user — for example, a tool that
+   * spawns a worker chat whose first turn carries a seed prompt.
+   */
+  Tool = 'tool',
+  /** A system-generated notification rather than a direct user message. */
```
connor4312, line 547: Do you think we should start expanding from the placeholder `origin: { kind: MessageKind };` to include the tool ID or agent metadata about the initiator? It's duplicative with the chat initiator in the current implementation, but it makes the message more self-identifying and we will certainly start to add more things there as we start thinking about inboxes and such

connor4312 (dismissed):

clients/rust/crates/ahp-types/src/state.rs
connor4312: I notice this didn't get generated correctly. But this should also be fixed if we move to a type union vs an inline object

types/channels-chat/state.ts
colbylwilliams: Haven't spent a lot of time thinking about what the shape would be, but the reason I added `_meta` to everything in #240 was to ensure I could add a `agentId` and `parentId` to every message so I could represent the agent->child agent hierarchy. I subsequently determined it would be better represented using the new multi-chat, but I definitely think there's something here...

- #240

clients/rust/crates/ahp-types/src/state.rs
colbylwilliams: Addressed — promoted the inline `origin: { kind: MessageKind }` to a named `MessageOrigin` interface in `types/channels-chat/state.ts` and registered it (plus `MessageKind`) in all four code generators. Rust now generates a proper `MessageOrigin` struct with a typed `kind: MessageKind` field instead of collapsing to `serde_json::Value`, and the same typed shape now surfaces in Go/Swift/Kotlin. Wire shape is unchanged.

docs/guide/state-model.md
```diff
@@ -199,6 +199,8 @@ MessageResourceAttachment {
 }
 ```
 
+A message's `origin.kind` records who initiated the turn: `user` for a direct user message, `agent` for a turn the agent seeds itself, `tool` for a turn a tool starts (for example, seeding the first turn of a worker chat it spawned), and `systemNotification` for a system-generated notification. A client is only allowed to send `user` messages.
```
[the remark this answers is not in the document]
colbylwilliams: Addressed — reworded the docs to describe the origin of the message itself (who produced it) rather than "who initiated the turn", with a note that for a turn-initiating message it is also the turn's origin.

connor4312 (approved):

PR #263, opened 2026-06-25
`root/progress` notification

Giuspepe:
## Why

Hosts fetch large native agent SDK/runtime binaries (70–95 MB) lazily on first use. Today that blocks for several seconds with nothing on the wire, so a client can only show a silent hang. This adds a generic host→client progress notification so clients can render a real progress indicator.

Related
- https://github.com/microsoft/vscode-internalbacklog/issues/8176
- https://github.com/microsoft/vscode/pull/322919

## Wire shape

A generic, **operation-agnostic** progress notification rather than a download-specific one. A client opts in per request by minting a `progressToken`; the host echoes it on every frame so the client can correlate progress back to the call (and the UI awaiting it) without the protocol naming a domain object.

**Opt-in** — new optional field on `createSession` (`CreateSessionParams`):

| field | notes |
|---|---|
| `progressToken?` | client-supplied token, offered up front. The host MAY ignore it (e.g. nothing long-running to do, SDK already cached), in which case no `root/progress` frames are emitted. Unique across the client's active requests. |

**Server→client** notification on the root channel, `root/progress` (`ProgressParams`):

| field | notes |
|---|---|
| `channel` | root channel URI |
| `progressToken` | echoes the token from the originating request |
| `progress` | progress so far in operation-defined units (e.g. bytes); monotonically non-decreasing |
| `total?` | present when known up front (e.g. `Content-Length`); absent ⇒ indeterminate |
| `message?` | optional human-readable message; clients that track the token render their own localized label |

```json
{ "jsonrpc": "2.0", "method": "root/progress",
  "params": { "channel": "ahp-root://",
              "progressToken": "9b2c1f7e-4a0d-4e2b-8b1a-2f7e4a0d4e2b",
              "progress": 18874368, "total": 41957498 } }
```

**Completion** is signalled by `progress === total` — no phase/error on the wire. The host emits a terminal frame with `total === progress` (when the total was never known, it sets `total` to the final `progress` on that frame); real failures surface via the existing session-failure path. Ephemeral (not replayed on reconnect); a client that misses the terminal frame should expire its indicator after an idle timeout.

The token is anchored on `createSession`, so it survives the provisional → materialized gap: the SDK download fires lazily when a session first materializes on its opening message, and the host fans the host-level download out to each waiting session's token.

Because the notification says nothing about *what* is progressing, the same channel can report any future long-running operation (other runtimes, plugins, models) without a new method.

Introduced in the unreleased 0.5.0 spec (`NOTIFICATION_INTRODUCED_IN`). Additive, non-breaking. Includes regenerated Rust/Kotlin/Swift/Go clients + JSON schema, and spec + per-client CHANGELOG entries.

### DIFF
--- a/types/channels-root/notifications.ts
+++ b/types/channels-root/notifications.ts
@@ -143,63 +143,41 @@ export interface SessionSummaryChangedParams {
   changes: Partial<SessionSummary>;
 }
 
-// ─── root/downloadProgress ───────────────────────────────────────────────────
+// ─── progress ────────────────────────────────────────────────────────────────
 
 /**
- * Lifecycle phase of a single download.
+ * Generic progress notification for a long-running operation.
  *
- * @category Protocol Notifications
- */
-export const enum DownloadPhase {
-  /** The download has begun; no bytes received yet. */
-  Started = 'started',
-  /** A throttled progress sample with bytes received so far. */
-  Progress = 'progress',
-  /** Terminal success frame; the resource is fully downloaded. */
-  Completed = 'completed',
-  /** Terminal failure frame; see {@link DownloadProgressParams.error}. */
-  Failed = 'failed',
-}
-
-/**
- * Broadcast on the root channel while the host downloads a resource on the
- * client's behalf — typically a multi-MB artifact fetched lazily the first time
- * it is needed (today: an agent's native SDK/runtime, `kind: 'agent-sdk'`). Lets
- * clients show a progress indicator instead of a silent multi-second hang.
+ * A client opts in to progress for a request by including a `progressToken` in
+ * that request (today: the `progressToken` field on `createSession`). If the
+ * server does long-running work to service the request — e.g. lazily
+ * downloading an agent's native SDK the first time a session of that provider
+ * is materialized — it emits `progress` notifications carrying the same token.
  *
- * The notification is intentionally **resource-agnostic** so the same channel
- * can report future downloads (additional agent runtimes, plugins, models, …)
- * without a new method. The `kind` discriminant categorizes the resource and
- * `resourceId` identifies it within that kind; clients that don't care can show
- * a single generic indicator driven by `displayName` + the byte counts.
- *
- * This is **host-level**, not session state: the artifact is shared across every
- * consumer and the host deduplicates concurrent fetches into one download (one
- * `downloadId`). The optional `session` field names the session whose action
- * triggered the fetch, purely as context — a client MAY attribute the progress
- * to that session's row, or show a single global indicator and ignore it.
+ * The notification is operation-agnostic: it says nothing about *what* is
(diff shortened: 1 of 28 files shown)


types/channels-root/notifications.ts
```diff
@@ -142,3 +142,121 @@ export interface SessionSummaryChangedParams {
...
+ *     "displayName": "Claude",
+ *     "phase": "progress",
+ *     "receivedBytes": 18874368,
+ *     "totalBytes": 41957498,
+ *     "session": "ahp-session:/<uuid>"
+ *   }
+ * }
+ * ```
+ */
+export interface DownloadProgressParams {
```
connor4312: Download progress is _very_ specific. I suggest we actually have a more generalized progress, maybe we borrow [from MCP](https://modelcontextprotocol.io/specification/2025-11-25/basic/utilities/progress) here. In their model progress messages are more generic (similary to our extension host progress) and can be associated with an RPC call, or not (in our case rather than including session in the progress, the progress would be associated with the 'create session' or subscribe call)

types/channels-root/notifications.ts
```diff
@@ -142,3 +142,121 @@ export interface SessionSummaryChangedParams {
...
+ * @category Protocol Notifications
+ * @method root/downloadProgress
+ * @direction Server → Client
+ * @messageType Notification
+ * @version 1
+ * @example
+ * ```json
+ * {
+ *   "jsonrpc": "2.0",
+ *   "method": "root/downloadProgress",
```
roblourens: What triggers a "download"? Is this really a global thing or is it blocking some other action that the client initiated?

I'm also interested in a more generic "progress" that can communicate a state like "Creating worktree...", I'm not sure whether that would fit with this or not

types/channels-root/notifications.ts
Giuspepe: thanks for the suggestion, made it more generic now

types/channels-root/notifications.ts
Giuspepe: E.g. when a client sends a message in a new session but that provider's agent SDK has not been downloaded yet, that triggers the server to download that agent SDK. This is for the Claude & Codex providers because their SDKs are not shipped with the product but are downloaded on demand. So we download them only when the client sends the first message in a new Claude/Codex session.

types/common/messages.ts
```diff
@@ -231,6 +232,7 @@ export interface ServerNotificationMap {
   'root/sessionAdded': { params: SessionAddedParams };
   'root/sessionRemoved': { params: SessionRemovedParams };
   'root/sessionSummaryChanged': { params: SessionSummaryChangedParams };
+  'progress': { params: ProgressParams };
```
connor4312: Let's define this as `root/progress` for routing

connor4312 (approved):

PR #265, opened 2026-06-25
sessions: add session-level inputNeeded aggregate

connor4312:
## Summary

Surface every outstanding input request across a session's chats on the **session channel**, so a client (mobile app, tool-providing client, etc.) can discover and answer them without subscribing to each chat.

This is **state-only** by design: each aggregate entry carries the owning `chat` URI plus all identifiers needed to respond, so a client answers by dispatching the ordinary `chat/*` action to that chat's channel — no chat subscription, no new response actions.

## What changed & why

- **`SessionState.inputNeeded`** — host-managed aggregate of outstanding input across all chats, keyed by an opaque `id`.
- **`SessionInputRequest`** union (discriminated on `kind`) with three variants:
  - `SessionChatInputRequest` — mirrors a chat's `ChatInputRequest` (elicitation).
  - `SessionToolConfirmationRequest` — a `ToolCallConfirmationState` awaiting confirmation, plus `turnId`.
  - `SessionToolClientExecutionRequest` — a running tool the session delegates to an active `clientId`.
- **`ToolCallConfirmationState`** union (`ToolCallPendingConfirmationState | ToolCallPendingResultConfirmationState`) — an inline anonymous union isn't generatable across clients, so it's named.
- **`session/inputNeededSet` / `session/inputNeededRemoved`** server actions (upsert / remove) + canonical reducer + 6 fixtures (keeps reducer branch coverage at 100%).
- **Cross-language**: wired the new types/unions/actions through all four code generators and hand-ported the reducer logic into the Rust, Go, Kotlin, and Swift clients.
- **Docs/CHANGELOGs**: new "Aggregated input requests" section in the session-channel spec; Unreleased entries in the root + all five client CHANGELOGs.

## Reviewer notes

- `SessionToolClientExecutionRequest.toolCall` is typed `ToolCallState` (documented as `running`) rather than `ToolCallRunningState`: a direct single-variant struct reference can't round-trip the `status` discriminant on the wire in the Rust/Go generators (the discriminant lives on the union wrapper). Happy to switch to a precise one-variant union if preferred.

## Testing

- Root suite: **273 tests, 100% reducer branch coverage**; lint + changelog checks pass.
- TS client (59), Rust, Go (build/vet/test), Swift (107) all pass. Generation is drift-free.
- **Kotlin not run locally** (no JDK in this environment); the port mirrors the verified TS/Rust/Go/Swift reducers with symbol names confirmed against generated output — relying on CI.


types/channels-session/state.ts
```diff
@@ -158,6 +180,136 @@ export interface SessionActiveClient {
...
+ *
+ * @category Session Input Types
+ */
+export const enum SessionInputRequestKind {
+  /** A user-facing elicitation mirrored from a chat's `inputRequests`. */
+  ChatInput = 'chatInput',
+  /** A tool call awaiting parameter- or result-confirmation. */
+  ToolConfirmation = 'toolConfirmation',
+  /** A running tool the session wants an active client to execute. */
+  ToolClientExecution = 'toolClientExecution',
```
roblourens, line 243: I don't think I'm following this part? Everything else makes sense

connor4312: An active client can be a participant of a session and provide its tools for all chats. But in order to know a chat needs is waiting for it to run a tool, it needs to be subscribed to that chat. This solves that case just like the confirmation/input request parts do

alexr00 (approved):

PR #270, opened 2026-06-26
Add capability stability levels to the initialize handshake

### LINKED ISSUE #269: Proposal: capability stability levels — let a host advertise experimental surface on the wire
## Summary

AHP's [Versioning](https://github.com/microsoft/agent-host-protocol/blob/main/docs/specification/versioning.md) page already defines a **"Capabilities First, Then Required"** lifecycle: new behavior lands as an opt-in capability, implementors check for it before use, and a later version may promote it to baseline. Today that signal is **binary** — a capability/surface is either advertised or it isn't — and additive changes "MUST be ignored by older peers that do not understand them."

There is a middle state with no vocabulary: a surface that is **implemented and emitted on the wire, but whose shape is not yet stable** and may change or be removed within an otherwise-compatible version range. This proposal adds a small, standard **stability dimension** to capability advertisement so a host can say "this exists, but it's experimental," and a client can decide how defensively to treat it — without inventing a per-host convention.

## Motivation

An agent host commonly wraps a fast-moving agent backend (a CLI, runtime, or remote service). Those backends frequently expose **experimental** surfaces — new event types, new fields — that they explicitly reserve the right to reshape. It's routine for an agent SDK's own schema to carry a per-method / per-type `stability` marker (`stable` / `experimental`) for exactly this reason; codegen surfaces it as `@experimental` doc annotations.

That stability information is known at the host's **build/schema** time but **does not currently cross the AHP wire**. A host relaying such a surface today has only two honest options:

1. **Drop it** — don't translate the experimental surface to AHP at all. Safe, but the client loses fidelity and can never adopt the feature early.
2. **Emit it silently** — translate it like any stable surface. The client gets the data but has no way to know it's volatile, so conservative clients either avoid it (back to option 1) or build on it and break when the shape churns.

Neither lets a client make an informed choice. A standard stability signal turns this into: *emit the surface, mark it experimental, and let the client opt in with eyes open.*

## Why this belongs in AHP (not a per-host convention)

Stability is a **generic protocol concern**: every host wrapping an evolving backend hits it, and every client benefits from handling it **uniformly** regardless of which host it's talking to. AHP already owns capability negotiation and the SemVer lifecycle, so this is an additive, idiomatic extension of an existing concept rather than a foreign one. If each host instead stamped its own vendor-namespaced metadata flag, clients would need per-host special-casing for what is fundamentally one cross-cutting idea.

## Why capability-level, not per-payload

Stability is a static property of a **surface** (a notification method, command, channel, or named capability) — not of an individual message instance. Every instance of an experimental notification is equally experimental. Stamping a flag onto every payload is redundant data that bloats the wire and invites drift between instances. Advertise it **once**, at the surface level, in the handshake.

## Proposed shape (illustrative — open to bikeshedding)

Advertise stability in the [`initialize`](https://github.com/microsoft/agent-host-protocol/blob/main/docs/specification/lifecycle.md) handshake result (and carry it through `reconnect`), as an optional map keyed by a surface identifier:

```jsonc
// InitializeResult (Server -> Client), new optional field
{
  "protocolVersion": "0.5.0",
  "serverSeq": 42,
  "snapshots": [ /* ... */ ],
  "featureStability": {
    // key: a surface identifier — notification method, command,
    // channel scheme, or named capability
    "session/canvasOpened":    "experimental",
    "session/scheduleRearmed": "experimental",
    "session/someOlderThing":  "deprecated"
    // anything not listed is "stable" by default
  }
}
```

Levels:

| Level | Meaning | Client guidance |
|---|---|---|
| `stable` (default / absent) | Normal SemVer additive guarantees apply. | Build on it freely. |
| `experimental` | On the wire, but shape may change or be removed within a compatible version range. | Parse defensively; gate behind a flag; tolerate shape changes; don't assume longevity. |
| `deprecated` | Still emitted, but scheduled for removal. | Migrate off; don't adopt anew. |

Notes / open questions:

- **Granularity.** Keyed by method/command name is the simplest. If finer granularity (a single experimental *field* on an otherwise-stable type) turns out to matter, that's arguably better left to the type's own documentation than the wire — worth deciding explicitly.
- **Agent-scoped vs connection-scoped.** Some surfaces may be agent-specific. The descriptor could also (or instead) hang off [`AgentInfo`](https://github.com/microsoft/agent-host-protocol/blob/main/docs/specification/root-channel.md) in the root channel. Connection-level is the simplest starting point.
- **Semantics are about *wire/schema stability*** ("may change or be removed"), explicitly **not** product gating, entitlement, or preview-access. Clients use it to decide how defensively to code, nothing more.
- **Relationship to "Capabilities First, Then Required."** This adds a *stability* axis to a capability: a surface can be advertised **and** flagged `experimental`, giving an explicit, observable path of `experimental -> stable -> (eventually) required` that the versioning page already gestures at.
- **Back-compat.** Optional and additive; older clients that ignore `featureStability` behave exactly as today (treat everything as stable), consistent with the existing "ignore what you don't understand" rule.

## Prior art

- **LSP / DAP capabilities** — feature presence negotiated at init.
- **MCP capabilities** — opt-in feature advertisement in the handshake.
- **OpenRPC / JSON Schema** — `x-`/annotation-style stability markers on methods and types.
- The common practice in agent SDK schemas of carrying a per-method/per-type `stability` field that codegen renders as `@experimental`.

## What I'm asking for

1. Is a **stability dimension on capability advertisement** something AHP wants to standardize (vs. leaving each host to convention)?
2. If so, preferred **attachment point** (`InitializeResult` vs `AgentInfo` vs a dedicated capabilities object) and **granularity** (surface-level vs field-level)?
3. Any appetite for the host-runtime piece — i.e. a host being able to *derive* these levels from a backend's existing schema `stability` markers, so the advertisement stays honest without hand-maintenance?

colbylwilliams:
Closes #269.

## What

Adds a standard **stability dimension** to capability advertisement, so a host can say "this surface exists, but it's experimental" and a client can decide how defensively to treat it — without inventing a per-host convention.

Concretely, this introduces:

- A `FeatureStabilityLevel` enum — `stable` / `experimental` / `deprecated`.
- An optional `InitializeResult.featureStability` map (`Record<string, FeatureStabilityLevel>`), keyed by a **surface identifier**: a notification method (e.g. `"session/canvasOpened"`), a command method (e.g. `"listSessions"`), a channel URI scheme (e.g. `"ahp-otlp:"`), or a named capability.

```jsonc
// InitializeResult (Server → Client)
{
  "protocolVersion": "0.5.0",
  "serverSeq": 42,
  "snapshots": [ /* ... */ ],
  "featureStability": {
    "session/canvasOpened": "experimental",
    "session/someOlderThing": "deprecated"
  }
}
```

| Level | Meaning | Client guidance |
|---|---|---|
| `stable` (default / absent) | Normal SemVer additive guarantees apply. | Build on it freely. |
| `experimental` | On the wire, but shape may change or be removed within a compatible version range. | Parse defensively; gate behind a flag; tolerate shape changes. |
| `deprecated` | Still emitted, but scheduled for removal. | Migrate off; don't adopt anew. |

## Design decisions (resolving the issue's open questions)

- **Attachment point → `InitializeResult`.** This matches the proposal's lead recommendation and the existing precedent of connection-level handshake metadata (`telemetry`, `defaultDirectory`, `completionTriggerCharacters`). Agent-scoped granularity on `AgentInfo` can be layered on later if a concrete need appears.
- **Granularity → surface-level** (keyed by method/command/scheme/capability name), not per-payload. Stability is a static property of a surface, so it's advertised once in the handshake rather than stamped on every message.
- **Default is `stable`.** A surface absent from the map carries the normal additive SemVer guarantees, so hosts list only the non-stable surfaces. The signal is strictly about **wire/schema volatility** — not product gating, entitlement, or preview access.
- **Carry through reconnect.** Rather than re-transmitting it (which would diverge from how other `InitializeResult` metadata behaves), the field is documented as connection-level metadata that the client **retains** across a `reconnect` — a reconnect continues the same logical connection. Older clients that don't understand the field ignore it and treat everything as stable, consistent with the existing "ignore what you don't understand" rule.

## Changes

- **`types/`** — new `FeatureStabilityLevel` enum + `InitializeResult.featureStability` field; exported from `index.ts`.
- **Generators** — registered the enum in `COMMAND_ENUMS` for the Go/Rust/Kotlin/Swift generators; regenerated all five clients, the JSON Schema, and the reference docs.
- **Spec docs** — new **Stability Levels** section in `versioning.md`; a note in the `initialize` response section of `lifecycle.md`.
- **CHANGELOGs** — `Added` entries under `[Unreleased]` in the spec changelog and all five client changelogs (a `types/` change ripples to every client).

## Validation

- `npm run typecheck`, `npm run lint`, `verify:changelog`, `verify:release-metadata` — pass.
- `npm run generate` is idempotent (clean tree after regeneration).
- Types unit tests: 267 pass.
- Per-client builds/tests: Go (build + test), Rust `ahp-types` (build + tests + doc-tests), TypeScript (typecheck + 59 tests), Swift (build), Kotlin (compile + tests) — all green. Confirms the `Record<string, FeatureStabilityLevel>` map-of-enum serializes correctly in every language.

> Note: `npm run test` fails locally only at the `c8` coverage wrapper (yargs is incompatible with the local Node v26); the underlying unit tests all pass when run directly. This is a pre-existing environmental issue unrelated to this change.

The protocol version is intentionally **not** bumped — the change is additive and lands under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,11 +25,12 @@ changes accumulate. Track in-flight protocol changes via PRs touching
 
 ### Added
 
+- `InitializeResult.featureStability` — an optional map advertising the
+  wire/schema stability (`FeatureStabilityLevel`: `stable` / `experimental` /
+  `deprecated`) of selected surfaces a host emits, keyed by surface identifier.
+  Surfaces absent from the map are `stable`.
 - `SessionModelInfo.maxOutputTokens` and `SessionModelInfo.maxPromptTokens`
   optional fields for communicating model token limits.
-- `SessionSummary._meta` optional provider metadata field for lightweight
-  session-list presentation hints.
-- `JsonPrimitive` type alias (`string | number | boolean | null`) in `types/common/state.ts`.
 - `session/activeClientRemoved` action to release a single active client from a
   session by `clientId`.
 
--- a/clients/go/CHANGELOG.md
+++ b/clients/go/CHANGELOG.md
@@ -16,6 +16,10 @@ tag whose matching `## [X.Y.Z]` heading is missing from this file.
 
 ### Added
 
+- `InitializeResult.FeatureStability` — an optional
+  `map[string]FeatureStabilityLevel` advertising the wire/schema stability
+  (`stable` / `experimental` / `deprecated`) of selected surfaces a host emits,
+  keyed by surface identifier. Surfaces absent from the map are `stable`.
 - `SessionModelInfo.MaxOutputTokens` and `SessionModelInfo.MaxPromptTokens`
   optional fields for communicating model token limits.
 - `SessionSummary.Meta` (wire `_meta`) optional provider metadata field for
--- a/clients/go/ahptypes/commands.generated.go
+++ b/clients/go/ahptypes/commands.generated.go
@@ -23,6 +23,30 @@ const (
 	ReconnectResultTypeSnapshot ReconnectResultType = "snapshot"
 )
 
+// Wire/schema stability of a surface advertised in
+// {@link InitializeResult.featureStability}.
+//
+// Stability describes how settled a surface's shape is within a compatible
+// SemVer range. It is deliberately **not** a product-gating, entitlement, or
+// preview-access signal — a client uses it only to decide how defensively to
+// code against the surface.
+type FeatureStabilityLevel string
+
+const (
+	// Normal additive SemVer guarantees apply; clients MAY build on it freely.
+	// This is the default for any surface absent from
+	// {@link InitializeResult.featureStability}, so it never needs to be
+	// advertised explicitly.
+	FeatureStabilityLevelStable FeatureStabilityLevel = "stable"
+	// On the wire, but its shape MAY change or be removed within an otherwise
+	// compatible version range. Clients SHOULD parse it defensively, gate
+	// adoption behind a flag, tolerate shape changes, and not assume longevity.
+	FeatureStabilityLevelExperimental FeatureStabilityLevel = "experimental"
+	// Still emitted, but scheduled for removal in a future version. Clients
+	// SHOULD migrate off it and SHOULD NOT adopt it anew.
+	FeatureStabilityLevelDeprecated FeatureStabilityLevel = "deprecated"
+)
+
 // Encoding of fetched content data.
 type ContentEncoding string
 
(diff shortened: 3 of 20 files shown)


connor4312:
I'm curious how you see a client would respond to a stability level here. As a client I would want to code 'defensively' for an experimental area, but there'd be no reason for me to have logic to remove that defensive code if the other side says their implementation of a feature is stable.

I was thinking that we'd probably add stability levels to new feature areas, but [similar to Node.js](https://nodejs.org/api/documentation.html#stability-index) these would be defined per feature area per spec version to make consumers aware of possible changes.

roblourens:
Same, not seeing why this is a runtime thing

PR #283, opened 2026-06-29
chat: add optional tool invocation intention

roblourens:
## Summary

- add an optional human-readable `intention` to `chat/toolCallStart` and every tool-call lifecycle state
- preserve the field through the TypeScript, Rust, Kotlin, Swift, and Go reducers
- regenerate schemas and client wire types, with shared fixture coverage and changelog entries

## Validation

- `npm run generate`
- `npm test`
- TypeScript tests and build
- Rust fmt, Clippy, and workspace tests
- Swift build and tests
- Kotlin Gradle build
- Go vet and tests

(Written by Copilot)


types/channels-chat/actions.ts
```diff
@@ -151,6 +151,8 @@ export interface ChatToolCallStartAction extends ToolCallActionBase {
   toolName: string;
   /** Human-readable tool name */
   displayName: string;
+  /** Human-readable description of what the tool invocation intends to do */
+  intention?: string;
```
connor4312, line 155: Do we also want some "risk" indicator at the same time so we can have the terminal intent classifier in AH? I would maybe dive into MCP's SEPs/discussions in this area to get some good prior art

roblourens: Yeah maybe, let's spec that when Christof is ready to get into it

connor4312 (approved):

PR #288, opened 2026-06-30
Complete the child-customization model (per-child enable/disable + invocation matrix)

### LINKED ISSUE #285: Complete the child-customization model: per-child enable/disable + a symmetric invocation matrix
## Summary

`SessionState.customizations` lets a host surface host-discovered **skills** and **custom agents** as native `SkillCustomization` / `AgentCustomization` children. In practice a host hits three gaps that force it to either *drop* entries or lose state, because the child-customization model is only half-built:

1. **No per-child enable/disable.** A skill or agent that exists but is individually turned off (config-level) has no representation. `enabled` lives only on container customizations, and `SessionCustomizationToggled` only reaches top-level containers — so an individually-disabled child can't be surfaced, only omitted.
2. **No user-invocation gate.** `SkillCustomization` has `disableModelInvocation` but no user-side counterpart, so "this skill is a user slash-command vs. model-auto-invoked" can't be expressed. A host that wants to render a command palette can't tell which entries are user-invocable.
3. **Asymmetric invocation model between Skill and Agent.** `AgentCustomization` has *neither* invocation gate, so "this agent is/isn't user-selectable" and "the model may/may not auto-delegate to it" are both unrepresentable.

The result is that disabled and non-user-invocable entries get dropped entirely — a settings UI can't list-and-re-enable a disabled skill, and a picker can't show (greyed out) the agents a user can't currently select.

## Current state in the spec (0.4.0)

- `ChildCustomization` variants (`Agent`, `Skill`, `Prompt`, `Rule`, `Hook`, `McpServer`) share `id` / `uri` / `name` / `icons` / `range`. None carries `enabled`.
- `SkillCustomization` carries `disableModelInvocation?: boolean`.
- `AgentCustomization` carries `_meta` but no invocation gates.
- `RuleCustomization` models activation with first-class, type-specific fields (`alwaysApply`, `globs`) — i.e. the spec's established style is to model behavior with real fields, not a generic `_meta` slot.
- Container customizations (`PluginCustomization`, `DirectoryCustomization`) carry `enabled: boolean`.
- The `SessionCustomizationToggled` reducer matches a **top-level container** id and sets the container's `enabled`; it has no path to toggle an individual child.

So the spec already has (a) an enable/disable *action* and (b) *one corner* of an invocation model (`Skill.disableModelInvocation`). This proposal finishes both.

### A. Per-child `enabled` + child-addressable toggle

- Add `enabled?: boolean` (absent ⇒ enabled) to the child customizations that can be individually disabled — at minimum `SkillCustomization` and `AgentCustomization`; ideally to the shared child base next to `id`/`uri`/`name` so it's uniform.
- Extend `SessionCustomizationToggled` (and the reducer) so a **child** `id` can be toggled, not just a container. Today the reducer only matches top-level containers.
- Define precedence explicitly: a disabled container disables its children regardless of each child's `enabled` (effective-enabled = container.enabled && child.enabled).

This lets a disabled-but-present customization be surfaced with accurate state instead of omitted, and generalizes cleanly — "toggle one skill/agent/rule without touching its siblings" is broadly useful.

> Alternative considered: mirror the container convention with a required `enabled: boolean` on children. An optional field with absent⇒enabled is friendlier for existing producers (no behavior change unless they set it). Either is fine; flagging the choice.

### B. A complete, symmetric invocation matrix on Skill **and** Agent

Add the missing gates so both child types carry the full {user, model} matrix, using the spec's existing negative-default-false convention (default = both can invoke):

| Field | `SkillCustomization` | `AgentCustomization` |
|---|---|---|
| `disableModelInvocation?: boolean` | exists | **add** (gates model auto-delegation as a sub-agent) |
| `disableUserInvocation?: boolean` | **add** (gates user slash-command invocation) | **add** (gates user selection in a picker) |

With (A) and (B), a host never has to drop an entry: a disabled skill is `enabled: false`; a non-user-invocable agent is `disableUserInvocation: true`; a model-only skill is `disableUserInvocation: true` (or a slash-only skill is `disableModelInvocation: true`). All are surfaced with faithful state.

## Why first-class fields, not `_meta`

`enabled` and user/model invocation are general agent-host concepts, not vendor-specific. The spec already models this class of thing with real fields (`Rule.alwaysApply`/`globs`, `Skill.disableModelInvocation`). Routing them through `_meta` would defeat the purpose of having native customization types and fragment the same concept across every vendor's `_meta` namespace. `_meta` should stay reserved for genuinely provider-specific residue.

## Concrete grounding (public, vendor SDK example)

A representative agent SDK already reports these per-entry on its discovery events, which is exactly the data with no AHP home:

- `session.skills_loaded` skill entries report `enabled` and `userInvocable` (and a model-invocation flag is being added — see the [companion SDK request](https://github.com/github/copilot-sdk/issues/1850)).
- `session.custom_agents_updated` agent entries report `userInvocable`, plus `model` and `tools`.

So the data exists at the source; only the AHP carrier fields are missing.

## Backwards compatibility

All additions are additive and optional (booleans / optional fields), so existing producers and consumers are unaffected until they opt in. The one behavioral extension is teaching `SessionCustomizationToggled` to match child ids; that only changes behavior for ids that previously matched nothing.

## Scope / related proposals

(A) and (B) are the core of this issue and hang together. Two adjacent improvements are tracked separately so each can land on its own:

- [Add first-class `model` and `tools` fields to `AgentCustomization`](https://github.com/microsoft/agent-host-protocol/issues/286) — behavior configuration (model override + tool scope).
- [Add a non-directory grouping container for host-discovered customizations](https://github.com/microsoft/agent-host-protocol/issues/287) — a `Customization::Group` container for non-file-backed groupings.

colbylwilliams:
Closes #285.

## What & why

The customization model could express *container*-level enable/disable but not per-child state, and its agent/skill invocation gating was asymmetric. This completes both halves.

### A. Per-child enable/disable + child-addressable toggling
- New non-exported `ChildCustomizationBase` (extends `CustomizationBase`) adds an optional `enabled` flag, shared by the five leaf children — `AgentCustomization`, `SkillCustomization`, `PromptCustomization`, `RuleCustomization`, `HookCustomization`. **Absent ⇒ enabled.**
- `session/customizationToggled` (and every client reducer) now matches `id` against top-level containers first, then the children inside every container, and sets that entry's `enabled`.
- Documented precedence: a child's effective state is `container.enabled && (child.enabled ?? true)`, so disabling a container still disables all its children.

### B. Symmetric user/model invocation matrix
- `SkillCustomization` gains `disableUserInvocation` (it already had `disableModelInvocation`).
- `AgentCustomization` gains both `disableModelInvocation` and `disableUserInvocation`.

All fields are optional and additive — backward compatible, so no protocol-version bump.

## Key design decision

The issue suggested putting `enabled?` on the shared base "next to id/uri/name". I initially did exactly that (on `CustomizationBase`) and validated the result — but the **client generators** (Rust/Go/Kotlin/Swift) inherit the base's *optional* declaration and ignore the `required` redeclaration on `McpServerCustomization`/containers, flipping their `enabled` to optional. The JSON-schema generator, by contrast, resolves the override correctly and keeps them required — so the clients and schema would **disagree** on the wire contract for containers and MCP servers.

A dedicated `ChildCustomizationBase` that only the five leaf children extend (McpServer and the containers keep their own required `enabled`) sidesteps the conflict entirely: every generator agrees, and this matches the issue's "shared child base" intent.

## Changed surface
- **Canonical types:** `types/channels-session/{state,actions,reducer}.ts`.
- **Conformance fixtures:** `225` (toggle a child by id; also exercises the top-level-McpServer and skip-McpServer-container branches) and `226` (no-op for unknown child id). Reducer branch coverage stays at 100%.
- **Generated mirrors:** schema + all five client type mirrors (regenerated; idempotent).
- **Hand-written client reducers:** Rust / Go / Kotlin / Swift each gain a `set*ChildEnabled` helper and a child-search fallback in the toggle path.
- **Docs:** `docs/guide/customizations.md` (Children, metadata table, Toggling) and `docs/guide/actions.md`; reference docs regenerated.
- **CHANGELOGs:** root + all five clients under `## [Unreleased]`.

## Verification
`npm run generate` (idempotent) and `npm run test` (typecheck, lint, verify:release-metadata, verify:changelog, 268 reducer tests @ 100% branch coverage) pass. The shared fixtures `225`/`226` were confirmed running and green in **all five** implementations: TypeScript, Rust (`cargo test`), Go (`go test ./...`), Kotlin (`./gradlew test`), and Swift (`swift test`).


types/channels-session/reducer.ts
```diff
@@ -168,12 +168,33 @@ export function sessionReducer(state: SessionState, action: SessionAction, log?:
...
+      if (topIdx >= 0) {
+        const updated = list.slice();
+        updated[topIdx] = { ...list[topIdx], enabled: action.enabled };
+        return { ...state, customizations: updated };
+      }
+      let changed = false;
+      const updated = list.map(container => {
+        if (container.type === CustomizationType.McpServer) {
+          return container;
+        }
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — fixed in edf8346. Rewrote the child-search fallback as an early-return `for` loop that scans in place and only copies the list (and the matched container's children) when an id actually matches; the no-op path now returns the original state with zero allocations. Same outputs for fixtures 060/061/062/225/226 (TS still 100% branch coverage; Kotlin build green).

clients/kotlin/src/main/kotlin/com/microsoft/agenthostprotocol/Reducers.kt
```diff
@@ -543,10 +553,25 @@ public fun sessionReducer(state: SessionState, action: StateAction): SessionStat
...
+                        if (childIdx < 0) container else {
+                            changed = true
+                            val newChildren = children.toMutableList()
+                            newChildren[childIdx] = withChildCustomizationEnabled(newChildren[childIdx], a.enabled)
+                            withCustomizationChildren(container, newChildren)
+                        }
+                    }
+                }
+                if (!changed) state else state.copy(customizations = updated)
             }
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — fixed in edf8346. Rewrote the child-search fallback as an early-return `for` loop that scans in place and only copies the list (and the matched container's children) when an id actually matches; the no-op path now returns the original state with zero allocations. Same outputs for fixtures 060/061/062/225/226 (TS still 100% branch coverage; Kotlin build green).

types/channels-session/actions.ts
```diff
@@ -272,22 +272,25 @@ export interface SessionCustomizationsChangedAction {
...
- * containers have an `enabled` flag; children are always active when
- * their container is enabled. Is a no-op when no matching container is
- * found.
+ * Targets either a top-level container (plugin or directory) or an
+ * individual child (a skill, agent, or other entry inside a container) by
+ * `id`, and sets that entry's `enabled` flag. Disabling a container still
+ * disables all of its children — the effective state of a child is
+ * `container.enabled && (child.enabled ?? true)` — so toggling a child
+ * only matters while its container is enabled. Is a no-op when no
+ * customization (container or child) has the given `id`.
```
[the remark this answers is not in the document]
colbylwilliams: Fixed in 211ad0c. Reworded the JSDoc to say it matches `id` against every top-level customization first — a plugin or directory container, or a bare top-level MCP server — then against container children, and dropped the "(container or child)" qualifier on the no-op clause. Regenerated the Go/Rust type mirrors and JSON schema from the updated JSDoc.

docs/guide/customizations.md
```diff
@@ -129,17 +131,17 @@ state.customizations
...
 {
   type: 'session/customizationToggled'
-  id: string         // container id
+  id: string         // container or child id
   enabled: boolean
 }
 ```
 
-Only containers (plugins and directories) have an `enabled` flag — children are always active when their container is enabled. The action is a no-op if no container has that id.
+Both containers and children carry an `enabled` flag. The reducer matches `id` against the top-level containers first, then the children inside every container, and sets that entry's `enabled`. A child's effective state is `container.enabled && (child.enabled ?? true)`, so disabling a container disables all of its children regardless of each child's own flag, and a child toggle only takes effect while its container is enabled. The action is a no-op if no container or child has that id.
```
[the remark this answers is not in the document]
colbylwilliams: Fixed in 211ad0c. The Toggling section now says the reducer matches `id` against every top-level customization first — plugins, directories, and bare top-level MCP servers — then the children inside every container, and is a no-op when no customization has that id. Also updated the intro sentence, the `// any customization id` code comment, and the action-table row for consistency.

connor4312:
I'm OK with this in general but I didn't implement it initially for because the Copilot SDK has no way to actually _do_ this. The SDK just takes `pluginDirectories` but doesn't allow enabling/disabling the constintuent parts. (Of course we could copy and rewrite plugins before giving them to the SDK to remove parts that are disabled but.) Do you know if this is a use case the SDK plans to support?

colbylwilliams:
> I'm OK with this in general but I didn't implement it initially for because the Copilot SDK has no way to actually _do_ this. The SDK just takes `pluginDirectories` but doesn't allow enabling/disabling the constintuent parts. (Of course we could copy and rewrite plugins before giving them to the SDK to remove parts that are disabled but.) Do you know if this is a use case the SDK plans to support?

@connor4312 it does now — you can disable individual skills and subagents by name, and enable/disable whole plugins. That's what backs the skill/plugin toggles in the desktop app's settings today.

What's still missing is on the AHP-surfacing side: our host reports those skills/agents to clients as customizations, and without a per-child enabled (or the user/model invocability gate) there's no way to represent one that's toggled off — so we have to drop it instead of showing it as present-but-off. This PR lets us surface that state, the same way the desktop app lists a skill that's turned off. Containers and MCP servers already carry enabled on main; this just makes the leaf children symmetric.

types/channels-session/state.ts
```diff
@@ -791,6 +791,35 @@ export interface DirectoryCustomization extends ContainerCustomizationBase {
...
+   *
+   * This flag is independent of the parent container's: the **effective**
+   * enabled state of a child is
+   * `container.enabled && (child.enabled ?? true)`, so a disabled container
+   * disables every child regardless of each child's own flag.
+   *
+   * A child is turned on or off by id with
+   * {@link SessionCustomizationToggledAction | `session/customizationToggled`}.
+   */
+  enabled?: boolean;
```
connor4312, line 820: This can just be in CustomizationBase

types/channels-session/state.ts
```diff
@@ -823,6 +852,18 @@ export interface AgentCustomization extends CustomizationBase {
...
+   * as a sub-agent; it can only be selected by the user. Absent or
+   * `false` means the agent may delegate to it.
+   */
+  disableModelInvocation?: boolean;
+  /**
+   * When `true`, the user cannot select this custom agent (for example,
+   * in a picker); it remains available for the agent to auto-delegate
+   * to. Absent or `false` means the user may select it.
+   */
+  disableUserInvocation?: boolean;
```
connor4312, line 866: I would kind of prefer an array like `audience: ['user', 'assistant']` which is [consistent with MCP](https://modelcontextprotocol.io/specification/2025-11-25/server/resources#annotations), but we already have `disableModelInvocation` below which more closely represents what is already in [skill frontmatter](https://open-plugins.com/agent-builders/components/skills#command-file-format). Wdyt?

colbylwilliams: I think `disableModelInvocation` is the most understandable for clients and host implementors, even though I do personally like the array better.

types/channels-session/state.ts
colbylwilliams: I actually tried this and backed it out — `ChildCustomizationBase` is doing real work.

The asymmetry is the catch: containers (`PluginCustomization`/`DirectoryCustomization`) and top-level `McpServerCustomization` carry `enabled` as **required** — they always have a known on/off state — while children make it **optional** (`enabled?`, absent = enabled) so a producer only sets it to surface a child that's turned off.

To move it onto `CustomizationBase` I'd put an optional `enabled?` there and re-declare it required on the containers/MCP server. That type-checks, and the JSON schema honors the narrowing — but the four client generators (Rust/Kotlin/Swift/Go) flatten `extends` and keep the *base* declaration, so every client mirror would emit `enabled` as optional and disagree with the schema. The only way to keep the mirrors consistent is to loosen `enabled` to optional on the containers and MCP server too, which relaxes their existing required contract.

So `ChildCustomizationBase` is the seam that keeps children optional without touching the required `enabled` on the containers/MCP. Happy to go the other way if we're fine making `enabled` optional (absent = enabled) uniformly — that'd let it live on the base cleanly.

connor4312: mm, fine

connor4312 (dismissed):

TylerLeonhardt (dismissed):

connor4312:
@copilot merge + fix conflicts. The current set of changes should go under the 0.5.2 unreleased heading

connor4312 (approved):

TylerLeonhardt (approved):

PR #290, opened 2026-06-30
Fix SnapshotState decoders dropping ChatState

friggeri:
## Why

`SnapshotState` is an untagged union; each client distinguishes its variants structurally by probing for distinctive required fields. Those probes still keyed off a `summary` field that was removed when `SessionState` was flattened (commit `80454fd`), breaking snapshot decoding:

- Swift and Kotlin never attempted `ChatState`, so chat snapshots failed to decode (Swift threw; Kotlin fell through to `Root`).
- Go probed for `summary` on both the session and chat branches, so both silently decoded as `Root`.

## Approach

Fixed the decoders in the generators (`scripts/generate-*.ts`) and regenerated the clients. Disambiguation now uses real required fields: session -> `lifecycle`, chat -> `turns`, terminal -> `content`, changeset -> `status` + `files`, resource-watch -> `root` + `recursive`, annotations -> `annotations`, root -> catch-all. Swift and Kotlin gained the missing `chat` branch. Rust already decoded correctly via its `#[serde(untagged)]` enum, so only its stale doc comment changed.

## Notes for reviewers

- Session is probed before annotations because `SessionState` carries an optional `annotations` summary that reuses the same key.
- Added a Go regression test (`TestSnapshotStateVariants`) round-tripping every variant.
- Verified locally: Go build + tests, Rust build, Swift build, and repo typecheck/lint/verify:changelog/type tests pass. Kotlin couldn't be compiled locally (no JVM); generated output verified by inspection.

### DIFF
--- a/clients/go/CHANGELOG.md
+++ b/clients/go/CHANGELOG.md
@@ -19,6 +19,14 @@ tag whose matching `## [X.Y.Z]` heading is missing from this file.
 - Optional `Intention` field on `ChatToolCallStartAction` and every tool-call
   lifecycle state.
 
+### Fixed
+
+- `SnapshotState.UnmarshalJSON` now decodes the `Chat` variant. Variant
+  disambiguation previously probed for the removed `summary` field (a leftover
+  from before `SessionState` was flattened), so chat and session snapshots both
+  fell through to the `Root` catch-all. Sessions are now matched on `lifecycle`
+  and chats on `turns`.
+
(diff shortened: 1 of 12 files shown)


friggeri:
@microsoft-github-policy-service agree company="Microsoft"

connor4312 (approved):

pwang347 (approved):

roblourens (approved):

PR #291, opened 2026-07-01
Add AgentCapabilities to AgentInfo

sandy081:
## Summary

Adds an optional `capabilities` field to `AgentInfo`, backed by a new `AgentCapabilities` interface, so clients can gate features (multi-chat, fork, sub-agent teams) via advertised capability flags instead of switching on the provider id. Absent flags default to unsupported.

```ts
export interface AgentCapabilities {
  /** Agent can host more than one concurrent chat per session. */
  supportsMultipleChats?: boolean;
  /** Agent can fork a chat from a turn. */
  supportsFork?: boolean;
}
```

## Motivation

vscode PR microsoft/vscode#323625 (merged) hand-edited a **generated** file — `src/vs/platform/agentHost/common/state/protocol/channels-root/state.ts` — to add this interface. That folder is auto-generated by `scripts/sync-agent-host-protocol.ts` from this repo, so the change must live here or the next sync overwrites it on vscode main. Upstreaming it here makes the vscode copy generated rather than hand-edited and closes the overwrite window.

## Changes

- `types/channels-root/state.ts`: new `capabilities?: AgentCapabilities` field on `AgentInfo` plus the `AgentCapabilities` interface.
- Registered `AgentCapabilities` in each per-language generator (`generate-go`, `generate-kotlin`, `generate-rust`, `generate-swift`).
- Regenerated all client bindings, JSON schema, and docs (`npm run generate`).
- CHANGELOG entries under `## [Unreleased]` → `### Added` in the root and every `clients/<lang>/CHANGELOG.md` (a types/ change ripples to all clients).

## Verification

`npm run generate` and `npm test` (typecheck + lint + verify:release-metadata + verify:changelog + reducer 100% branch coverage) both pass.

## Follow-up

After this merges, re-sync the vscode side (`npx tsx scripts/sync-agent-host-protocol.ts` from the vscode checkout) and confirm `channels-root/state.ts` regenerates the same `AgentCapabilities` block.

### DIFF
--- a/types/channels-root/state.ts
+++ b/types/channels-root/state.ts
@@ -85,6 +85,25 @@ export interface AgentInfo {
    * into the session's `customizations` list.
    */
   customizations?: Customization[];
+  /**
+   * Static capability flags the agent advertises about itself. Clients use
+   * these to gate features (multi-chat, fork, sub-agent teams) instead of
+   * switching on the provider id. Absent flags default to unsupported.
+   */
+  capabilities?: AgentCapabilities;
+}
+
+/**
+ * Static capability flags an {@link AgentInfo} advertises. Each flag is opt-in
+ * (absent -> unsupported).
+ *
+ * @category Root State
+ */
+export interface AgentCapabilities {
+  /** Agent can host more than one concurrent chat per session. */
+  supportsMultipleChats?: boolean;
+  /** Agent can fork a chat from a turn. */
+  supportsFork?: boolean;
 }
 
 /**
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ changes accumulate. Track in-flight protocol changes via PRs touching
   to do.
 - Optional `model` and `tools` fields on `AgentCustomization`, giving a custom
   agent's pinned model and tool allowlist a first-class home instead of `_meta`.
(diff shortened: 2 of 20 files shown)


sandy081:
Superseded: recreating from the correct worktree branch.

types/channels-root/state.ts
```diff
@@ -85,6 +85,25 @@ export interface AgentInfo {
...
+ * (absent -> unsupported).
+ *
+ * @category Root State
+ */
+export interface AgentCapabilities {
+  /** Agent can host more than one concurrent chat per session. */
+  supportsMultipleChats?: boolean;
+  /** Agent can fork a chat from a turn. */
+  supportsFork?: boolean;
 }
```
connor4312, line 107: 1. We should be a bit more explicit about what these mean for a UI. I.e. `supportsMultipleChats` means a client must not call `createChat` (creating chats beyond the default) and supportsFork means no `ChatForkSource`
2. Let's model this structure a bit more after MCP since we already speak MCP capabilities a bit here. In that world it would be

```

export interface AgentCapabilities {
  multipleChats?: {
    fork?: boolean;
  }
}
```

(so a client must support multiple chats for forking, and can give `multipleChats: {}` for just multi chat but no forking, or `multipleChats: { fork: true }`)

PR #295, opened 2026-07-01
Add cursor-based pagination to listSessions

### LINKED ISSUE #275: Proposal: paginate listSessions for large session catalogs
## Summary

Add optional **cursor + limit pagination** to `listSessions` so a client with a large session catalog can load it incrementally instead of receiving every `SessionSummary` in a single response. Fully additive: callers that omit the new fields, and hosts that do not paginate, behave exactly as today.

## Motivation

`ListSessionsParams` is `{ channel, filter? }` and `ListSessionsResult` is `{ items: SessionSummary[] }` — the whole catalog, in one response. A user with hundreds or thousands of sessions hands a constrained client one large payload to fetch, decode, and hold; and over a transport with a per-frame size ceiling, a big enough catalog risks the same frame-pressure that large snapshots face.

The session list is typically the **first thing a client paints**, so it should be pageable like any other potentially-large collection. Chat history already has incremental loading via `fetchTurns`; the session catalog — the tier above it — has no equivalent.

## Why this belongs in AHP

`listSessions` is the protocol's catalog entry point, and how a large catalog is delivered is a protocol concern. Pagination is the idiomatic, additive answer, and it keeps clients portable — a per-host paging convention would fragment them. It mirrors the shape AHP already chose for `fetchTurns` (an opaque backward cursor plus a "more exists" signal).

## Proposed shape (illustrative — open to bikeshedding)

```ts
export interface ListSessionsParams extends BaseParams {
  channel: 'ahp-root://';
  filter?: object;
  /** Max entries to return; the server MAY impose its own cap. */
  limit?: number;
  /** Opaque cursor from a previous result; omit for the first page. */
  cursor?: string;
}

export interface ListSessionsResult {
  items: SessionSummary[];
  /** Opaque cursor for the next page; absent ⇒ no more pages. */
  nextCursor?: string;
}
```

- The cursor is **opaque and server-defined** — ordering and keyset are the host's choice — matching how `fetchTurns` exposes `before` / `hasMore` without dictating storage.
- **Fully additive.** A client that omits `limit` / `cursor` and ignores `nextCursor` sees today's behavior (subject to any server-imposed cap); a host that does not paginate ignores the inputs and returns everything.
- **Ordering.** The server SHOULD return most-recently-modified first, so the first page is the useful one — consistent with the newest-first rendering pattern constrained clients want.

## Interaction with existing work

- The `root/sessionAdded` / `root/sessionRemoved` / `root/sessionSummaryChanged` notifications keep an already-loaded page live exactly as today; pagination governs only the initial and backfill fetches, not the live edge.
- Pairs with the `ChatSummary` preview proposal: pagination bounds *how many* rows a client fetches; preview makes *each* row renderable without a subscription. Together they make a large catalog affordable end-to-end on a small client.
- Composes with **#264** (which makes `SessionSummary` a root-only catalog type) — pagination is a property of the root catalog query and is unaffected by where the summary's fields live.

## Open questions

- **Cursor vs offset.** An opaque keyset cursor (recommended) avoids duplicates and skips when the catalog mutates mid-scroll; offset/limit is simpler but drift-prone under concurrent add/remove. Worth stating the guarantee the cursor provides (stable forward iteration).
- **Filter typing.** `filter` is an untyped `object` today. Suggest keeping it orthogonal — don't block paging on specifying the filter shape.
- **Total count.** Should the result optionally carry an approximate total (`approximateTotal?: number`) for "N sessions" UI, or is that over-reach for v1?

connor4312:
## Summary

Adds optional **cursor + limit pagination** to the root `listSessions` command so a client with a large session catalogue can load it incrementally instead of receiving every `SessionSummary` in a single response. Modeled on [MCP's opaque-cursor pagination](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/pagination), plus an optional `limit` the server **SHOULD** respect (but **MAY** cap).

Fully additive: a caller that omits the new fields, and a host that does not paginate, behave exactly as today.

Closes #275.

## Shape

The pagination inputs/outputs are factored into reusable base interfaces in `types/common/commands.ts`, so future paginated commands can just extend them and the pagination contract lives in one place:

```ts
export interface PaginatedParams {
  limit?: number;   // server SHOULD respect; MAY cap
  cursor?: string;  // opaque; omit for the first page
}

export interface PaginatedResult {
  nextCursor?: string; // absent ⇒ no more pages
}

export interface ListSessionsParams extends BaseParams, PaginatedParams {
  channel: 'ahp-root://';
  filter?: object;
}

export interface ListSessionsResult extends PaginatedResult {
  items: SessionSummary[];
}
```

- The cursor is **opaque and server-defined** — ordering and keyset are the host's choice, mirroring how `fetchTurns` exposes `before`/`hasMore` without dictating storage. Clients MUST NOT parse, modify, or persist it; an unrecognised cursor SHOULD be rejected with `InvalidParams`.
- The server SHOULD return **most-recently-modified first**, so the first page is the immediately useful one.
- The `root/session*` notifications keep an already-fetched page live exactly as today; pagination governs only the initial and backfill fetches.

## Changes

- **`types/`** — new `PaginatedParams` / `PaginatedResult`; `ListSessionsParams` / `ListSessionsResult` extend them; both exported from `types/index.ts`.
- **Generators** — added the two base interfaces to the `knownSpecial` exhaustiveness lists in the rust/go/kotlin/swift generators (like `BaseParams`, they're flattened via `extends`, not emitted standalone).
- **Regenerated** — JSON Schema and all five client mirrors (Rust, Kotlin, Swift, TypeScript, Go); the native clients flatten `limit`/`cursor`/`nextCursor` into `ListSessions*`.
- **Docs** — a "Paginating the catalogue" section in `docs/specification/root-channel.md`.
- **CHANGELOGs** — an entry in the spec changelog plus every client changelog (a `types/**` change ripples to all clients).

## Open questions from the issue (deliberately out of scope)

- **Sorting/filtering** (raised by @roblourens): kept orthogonal — `filter` stays an untyped `object` as today. Pagination is additive so a typed filter/sort can land later without reworking this.
- **`approximateTotal`**: omitted for v1.

## Validation

- `npm run test` → 275 pass, 100% reducer branch coverage, changelog + release-metadata gates green; `npm run lint` clean.
- `npm run generate` produces no generator warnings.
- `cargo build`, `go build ./...`, and the TypeScript client `tsc --noEmit` all pass.

### DIFF
--- a/types/channels-root/commands.ts
+++ b/types/channels-root/commands.ts
@@ -7,7 +7,7 @@
  */
 
 import type { URI } from '../common/state.js';
-import type { BaseParams } from '../common/commands.js';
+import type { BaseParams, PaginatedParams, PaginatedResult } from '../common/commands.js';
 import type { SessionSummary, SessionConfigSchema } from '../channels-session/state.js';
 
 // Re-export schema types so the legacy `commands.ts` aggregator continues to
@@ -24,21 +24,47 @@ export type { SessionConfigPropertySchema, SessionConfigSchema } from '../channe
  * large. Clients fetch it imperatively and maintain a local cache updated by
  * `root/sessionAdded` and `root/sessionRemoved` notifications.
  *
+ * A large catalogue can be fetched incrementally via the {@link PaginatedParams}
+ * `limit`/`cursor` inputs (see that type for the full pagination contract). The
+ * server SHOULD return most-recently-modified entries first, so the first page
+ * is the immediately useful one. The `root/session*` notifications keep an
+ * already-fetched page live; pagination governs only the initial and backfill
+ * fetches.
+ *
  * @category Commands
  * @method listSessions
  * @direction Client → Server
  * @messageType Request
  * @version 1
+ * @example
+ * ```jsonc
+ * // Client → Server (fetch the first page of up to 50 sessions)
+ * { "jsonrpc": "2.0", "id": 4, "method": "listSessions",
+ *   "params": { "channel": "ahp-root://", "limit": 50 } }
+ *
+ * // Server → Client (a cursor signals more entries exist)
+ * { "jsonrpc": "2.0", "id": 4, "result": {
+ *   "items": [ { "id": "s1", ... }, { "id": "s2", ... } ],
+ *   "nextCursor": "eyJvIjo1MH0="
+ * }}
+ *
+ * // Client → Server (fetch the next page)
+ * { "jsonrpc": "2.0", "id": 5, "method": "listSessions",
+ *   "params": { "channel": "ahp-root://", "limit": 50, "cursor": "eyJvIjo1MH0=" } }
+ * ```
  */
-export interface ListSessionsParams extends BaseParams {
+export interface ListSessionsParams extends BaseParams, PaginatedParams {
   channel: 'ahp-root://';
   /** Optional filter criteria */
   filter?: object;
 }
 
 /** Result of the `listSessions` command. */
-export interface ListSessionsResult {
-  /** The list of session summaries. */
+export interface ListSessionsResult extends PaginatedResult {
+  /**
+   * The list of session summaries. The server SHOULD order them
+   * most-recently-modified first.
+   */
   items: SessionSummary[];
 }
 
--- a/types/channels-root/commands.ts
+++ b/types/channels-root/commands.ts
@@ -55,8 +55,6 @@ export type { SessionConfigPropertySchema, SessionConfigSchema } from '../channe
  */
 export interface ListSessionsParams extends BaseParams, PaginatedParams {
   channel: 'ahp-root://';
-  /** Optional filter criteria */
-  filter?: object;
 }
 
 /** Result of the `listSessions` command. */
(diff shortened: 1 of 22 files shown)


types/channels-root/commands.ts
```diff
@@ -24,21 +24,47 @@ export type { SessionConfigPropertySchema, SessionConfigSchema } from '../channe
...
+ * // Client → Server (fetch the next page)
+ * { "jsonrpc": "2.0", "id": 5, "method": "listSessions",
+ *   "params": { "channel": "ahp-root://", "limit": 50, "cursor": "eyJvIjo1MH0=" } }
+ * ```
  */
-export interface ListSessionsParams extends BaseParams {
+export interface ListSessionsParams extends BaseParams, PaginatedParams {
   channel: 'ahp-root://';
   /** Optional filter criteria */
   filter?: object;
```
roblourens: How is `filter` interpreted? Should we add a sort criteria?

roblourens (dismissed):

types/channels-root/commands.ts
connor4312: Yea, idk, I forgot this was a thing. I'll just remove it for now

dmitrivMS (approved):

Issue #297, opened 2026-07-02
Proposal: Canvas — declared, instanced, agent-openable UI surfaces on a dedicated channel


colbylwilliams:
## Summary

AHP has no Canvas concept on the wire. There is no state slot, no action family, no method, and no capability flag for it — `grep -ri canvas` against `main` returns nothing in `types/`, `docs/`, `schema/`, or any language client.

Several JSON-RPC agent runtimes ship a **Canvas** surface: declarative, interactive UI surfaces (an editor, a live browser, a terminal, or arbitrary extension-authored HTML) that the agent can open, drive, and close through a small set of tools, and that a client renders in a side panel. A runtime fronted by an AHP client has nowhere to put those surfaces today — the agent's "open a canvas" call reaches the runtime and then dies, because AHP gives it no way to reach the renderer.

This proposal adds Canvas to AHP by **following the two patterns AHP already uses for exactly this class of problem**:

1. The **[MCP Apps](https://github.com/microsoft/agent-host-protocol/blob/main/docs/guide/mcp.md#mcp-apps)** pattern — a client **capability opt-in** plus a **small discovery surface** on session state — for "here is an opaque, client-rendered UI surface and here is what the host will do for it."
2. The **stateful per-resource channel** pattern used by [terminals](https://github.com/microsoft/agent-host-protocol/blob/main/docs/specification/terminal-channel.md), [changesets](https://github.com/microsoft/agent-host-protocol/tree/main/types/channels-changeset), and the [resource-watch channel](https://github.com/microsoft/agent-host-protocol/blob/main/docs/specification/resource-watch-channel.md) — one channel per open instance, carrying that instance's live state and its lifecycle traffic.

Crucially, the runtime→provider callbacks that back a canvas (`open` / `invoke action` / `close`) are **request/response**. AHP already models bidirectional request/response: the `resource*` family and `createResourceWatch` appear in both `CommandMap` and `ServerCommandMap` and may be initiated by either peer (see [`types/common/messages.ts`](https://github.com/microsoft/agent-host-protocol/blob/main/types/common/messages.ts)). Canvas provider callbacks map onto that existing machinery directly, instead of being simulated with correlated state.

[...]

PR #298, opened 2026-07-02
Add Canvas channel (spec + all clients)

colbylwilliams:
Implements the Canvas channel proposed in #297. A **canvas** is an opt-in, host-driven interactive UI surface — a document or spreadsheet editor, a live preview, a data dashboard, a design surface — that the agent can open alongside a session and drive over AHP, following the "one channel per resource" model already used by terminals and changesets.

The PR adds the full protocol surface in the canonical `types/` tree, wires it through the per-language generators, hand-ports the reducer to all five client mirrors, and documents it in the guide, specification, and generated reference.

## Protocol surface

- **Capability** — an opt-in `ClientCapabilities.canvas` flag; clients that omit it see no canvas surface.
- **Discovery** — `SessionState.canvases` (registry of openable canvases) and `SessionState.openCanvases` (catalogue of open instances), republished atomically via `session/canvasesChanged` / `session/openCanvasesChanged`. Clients contribute canvases through `SessionActiveClient.canvasProviders`. Discovery types: `SessionCanvasDeclaration`, `ClientCanvasDeclaration`, `SessionCanvasOperation`, `OpenCanvasRef`, `CanvasProviderSource` (`CanvasProviderKind` server/client), and `CanvasAvailability` (ready/stale).
- **Per-instance channel** — one `ahp-canvas:/<id>` channel per open instance, carrying `CanvasState`. The channel URI is the instance's sole identity.
- **Actions** — `canvas/updated` (sparse-merge presentation state) and `canvas/closeRequested` (client→host signal with a no-op reducer).
- **Provider requests** — `canvasOpen` / `canvasInvokeOperation` / `canvasClose`, registered in `ServerCommandMap` and mirrored in `CommandMap` (the `resource*` precedent). All target the instance's channel URI; the host resolves it back to the provider, or handles a server-side provider internally.
- **Content** — state-driven: renderers resolve `CanvasState.contentUri`, either loading a directly-reachable address or fetching it over the general `resourceRead` command, so canvas content works over a relay with no direct host connection.
- **Error** — `CanvasProviderError` (`-32012`) carrying `{ code, message }`.
- **Version** — canvas registers at the current `PROTOCOL_VERSION` (`0.6.0`); client packages stay `0.5.2` (bumped at release time). A changelog fragment lands under `docs/.changes/`.

## Cross-language surface

- **Types** authored in `types/` (capability, `channels-canvas/`, `channels-session/` discovery state + actions + reducer, common actions union, messages/version registries, errors, index).
- **Generators** wired through find-protocol-sources, generate-action-origin, and generate-{go,rust,kotlin,swift}; generate-markdown gets the canvas slug + reference page.
- **Clients** — reducer hand-ports for Rust, Kotlin, Swift, and Go plus the TS reducer, all exercised by the shared conformance fixtures.

## Fixtures

Reducer fixtures cover provider-registry replace, open-catalogue replace, `canvas/updated` sparse merge (present overwrites / absent preserves), and `canvas/closeRequested` no-op.

## Documentation

- **Guide** — `docs/guide/canvases.md`: the two client roles, discovery, state-driven rendering, availability, interaction, lifecycle, and worked mermaid flows.
- **Specification** — `docs/specification/canvas-channel.md`.
- **Reference** — generated `docs/reference/canvas.md`, plus canvas listed in the specification overview, the what-is-AHP glossary, and the reference sidebar.

### Rendered preview

GitHub renders these from the PR branch, mermaid diagrams included:

- [Canvases guide](https://github.com/microsoft/agent-host-protocol/blob/colby-canvas-channel/docs/guide/canvases.md)
- [Canvas Channel specification](https://github.com/microsoft/agent-host-protocol/blob/colby-canvas-channel/docs/specification/canvas-channel.md)

After merge to `main`, the site publishes them (the reference page is built there — it's generated, not committed):

- https://microsoft.github.io/agent-host-protocol/guide/canvases
- https://microsoft.github.io/agent-host-protocol/specification/canvas-channel
- https://microsoft.github.io/agent-host-protocol/reference/canvas

## Design decisions (deviations from the proposal)

1. **The channel URI is the sole instance identity.** Following `createTerminal` / `invokeChangesetOperation`, `canvasOpen` mints the instance's own `ahp-canvas:/<id>` channel and every subsequent request targets that URI — there is no separate `instanceId`.
2. **`canvas/updated` is a value-based sparse merge** (present overwrites, absent preserves) rather than the proposal's three-state `null`-clears model, which isn't portable across the generated Rust/Kotlin/Swift/Go nullable types or the conformance harness. A full reset arrives as a fresh `CanvasState` snapshot on (re)subscribe.
3. **Content is state-driven via `contentUri` + `resourceRead`**, not a bespoke in-band message path, so canvas content flows over the existing transport and works over a relay.
4. **`CanvasProviderSource` is a named discriminated union** (`CanvasProviderKind` + server/client variants) per the editorial `*Kind` rule and the generators' need for named variants.


CHANGELOG.md
```diff
@@ -23,6 +23,10 @@ changes accumulate. Track in-flight protocol changes via PRs touching
 `NOTIFICATION_INTRODUCED_IN` maps in
 [`types/version/registry.ts`](types/version/registry.ts).
 
+## [0.6.0] — Unreleased
+
+Spec version: `0.6.0`
+
```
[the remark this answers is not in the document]
colbylwilliams: Agreed — good catch, and reverted in 7f11f60. A feature PR shouldn't bump the spec version: per AGENTS.md and the `session/inputNeeded` precedent (ea386ee), new actions register at the *current* `PROTOCOL_VERSION` and their bullets accumulate under `## [Unreleased]`, while the version bump + `## [X.Y.Z]` rotation is a dedicated release chore (e.g. 90bd16a). So I reverted `PROTOCOL_VERSION` 0.6.0 → 0.5.1 (and dropped 0.6.0 from `SUPPORTED_PROTOCOL_VERSIONS`), registered the five canvas actions at 0.5.1, and folded the canvas bullet back under `## [Unreleased]`.

types/reducers.ts
```diff
@@ -12,4 +12,5 @@ export { terminalReducer } from './channels-terminal/reducer.js';
 export { changesetReducer } from './channels-changeset/reducer.js';
 export { annotationsReducer } from './channels-annotations/reducer.js';
 export { resourceWatchReducer } from './channels-resource-watch/reducer.js';
+export { canvasReducer } from './channels-canvas/reducer.js';
 export { softAssertNever, isClientDispatchable } from './common/reducer-helpers.js';
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — extended `isClientDispatchable` in `types/common/reducer-helpers.ts` to accept `CanvasAction` and narrow to `ClientCanvasAction`, so all three client-dispatchable canvas actions (`canvas/updated`, `canvas/closeRequested`, `canvas/message`) validate/narrow without a cast. Added a canvas assertion to its test. Fixed in 7f11f60.

docs/specification/canvas-channel.md
```diff
@@ -0,0 +1,208 @@
...
+```mermaid
+sequenceDiagram
+    autonumber
+    participant A as Agent
+    participant H as Host (server)
+    participant C as Client (provider + renderer)
+
+    Note over C: declares ClientCapabilities.canvas<br/>publishes canvasProviders
+    C->>H: dispatchAction session/activeClientsChanged (canvasProviders)
+    H->>C: action session/canvasesChanged (registry)
```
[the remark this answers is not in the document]
colbylwilliams: Already addressed in c5ff126 (this comment is anchored to an earlier commit) — the sequence diagram now uses `session/activeClientSet`.

docs/specification/canvas-channel.md
```diff
@@ -0,0 +1,208 @@
...
+    C-->>H: { url, title }
+    H->>C: action session/openCanvasesChanged (instance added)
+    C->>H: subscribe { channel: "ahp-canvas:/abc" }
+    H-->>C: snapshot { state: CanvasState }
+
+    Note over C: renderer loads CanvasState.url
+    A->>H: invoke action "format"
+    H->>C: canvasInvokeAction { instanceId, actionName, input }
+    C-->>H: { value }
+    C-->>H: action canvas/updated { status }
```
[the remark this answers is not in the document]
colbylwilliams: This one is now stale. Following earlier review feedback (c5ff126), `canvas/updated` was made `@clientDispatchable`: a client-side provider needs it to push its own presentation changes into `CanvasState`, mirroring how `terminal/titleChanged` lets a client-owned terminal rename itself. So the `C-->>H` (client→host) direction is intentional and matches the updated action table, which lists `canvas/updated` as “client → host (a client-side provider) **or** host → client.”

types/version/registry.ts
```diff
@@ -151,6 +151,11 @@ export const ACTION_INTRODUCED_IN: { readonly [K in StateAction['type']]: string
   [ActionType.TerminalCommandExecuted]: '0.1.0',
   [ActionType.TerminalCommandFinished]: '0.1.0',
   [ActionType.ResourceWatchChanged]: '0.2.0',
+  [ActionType.SessionCanvasesChanged]: '0.5.1',
+  [ActionType.SessionOpenCanvasesChanged]: '0.5.1',
+  [ActionType.CanvasUpdated]: '0.5.1',
+  [ActionType.CanvasCloseRequested]: '0.5.1',
+  [ActionType.CanvasMessage]: '0.5.1',
```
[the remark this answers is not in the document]
colbylwilliams: Good catch — the PR description was stale. I reverted the `PROTOCOL_VERSION` bump (itself in response to earlier review feedback), so the canvas actions now register at the current spec version `0.5.1` in `ACTION_INTRODUCED_IN`, with the CHANGELOG bullet under `## [Unreleased]`. The description has been updated to match — no `0.6.0` bump in this PR.

schema/notifications.schema.json
```diff
@@ -4571,6 +4771,64 @@
...
+          "description": "Renderer-targeted address for the opaque canvas content — either a\ndirectly-loadable URL (`https:`, an in-process scheme, `http://localhost`)\nor a channel-served `ahp-canvas-content:/<instanceId>/<path>` address the\nrenderer resolves over this channel with `canvasReadResource`. The\nrenderer dispatches on the scheme and enforces its URL policy. See\n{@link /specification/canvas-channel | Canvas Channel}."
+        },
+        "availability": {
+          "$ref": "#/$defs/CanvasAvailability",
+          "description": "Whether this instance's provider is currently available."
+        },
+        "provider": {
+          "$ref": "#/$defs/CanvasProviderSource",
+          "description": "Which provider owns the callbacks (`canvasOpen` / … ) for this instance."
+        }
```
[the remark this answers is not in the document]
colbylwilliams: This is a pre-existing, repo-wide limitation of the notifications schema generator rather than anything canvas-specific. `notifications.schema.json` references ~95 `$defs` it never inlines (e.g. `ToolCallState`, `SessionInputRequest`, `MessageAttachment`, `URI`, `JsonPrimitive`, and every `*Kind` enum member) because the generator only emits the interfaces it directly walks and does not transitively pull in referenced type aliases/enums. `CanvasState.provider → CanvasProviderSource` follows that identical established pattern. Fixing it means teaching the generator to transitively resolve/inline refs — out of scope for this feature PR, so I filed #302 to track it (along with the `oneOf` issue below).

schema/state.schema.json
```diff
@@ -4357,6 +4615,18 @@
...
+    "CanvasProviderSource": {
+      "oneOf": [
+        {},
+        {
+          "$ref": "#/$defs/CanvasServerProviderSource"
+        },
+        {
+          "$ref": "#/$defs/CanvasClientProviderSource"
+        }
+      ],
```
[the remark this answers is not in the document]
colbylwilliams: Also a pre-existing, repo-wide generator behavior. Every union *type alias* comes out with a leading empty `{}` branch — 15 `$defs` in `state.schema.json` (14 on `main` before canvas) and 17 in `actions.schema.json`, including `ToolCallState`, `SessionInputRequest`, `JsonPrimitive`, `ChatOrigin`, `MessageAttachment`, `ResponsePart`, etc. The cause is in shared logic: a union alias's `getText()` carries a leading `|`, so `splitUnionType` yields an empty-string first element that maps to `{}` (the handler filters `'undefined'` but not `''`). `CanvasProviderSource` reproduces the exact same pattern. The one-line fix regenerates all ~30 unrelated unions, so it does not belong in the canvas PR — tracked in #302.

types/channels-session/state.ts
```diff
@@ -221,9 +244,182 @@ export interface SessionActiveClient {
...
+ *
+ * @category Canvas Types
+ */
+export interface ClientCanvasDeclaration {
+  /** Provider-local canvas id, unique within the publishing client. */
+  canvasId: string;
+  /** Human-readable canvas name. */
+  displayName: string;
+  /** Human-readable description of the canvas. */
+  description: string;
```
connor4312, line 391: Is this needed for a canvas the client publishes? The server isn't going to be consuming this most likely right?

types/channels-session/state.ts
```diff
@@ -221,9 +244,182 @@ export interface SessionActiveClient {
...
+   * The instance's channel URI (`ahp-canvas:/<id>`). Subscribe to it to load
+   * the full {@link CanvasState}.
+   */
+  channel: URI;
+  /** Provider-local canvas id this instance was opened from. */
+  canvasId: string;
+  /** Owning provider id. */
+  extensionId: string;
+  /** Human-readable provider name. */
+  extensionName?: string;
```
connor4312: What do these mean in the context of AHP? It doesn't know about extensions. We do know about plugins and customizations. Is that the right place for this?

types/channels-session/state.ts
```diff
@@ -221,9 +244,182 @@ export interface SessionActiveClient {
...
+export interface OpenCanvasRef {
+  /** Server-assigned instance handle, unique within the session. */
+  instanceId: string;
+  /**
+   * The instance's channel URI (`ahp-canvas:/<id>`). Subscribe to it to load
+   * the full {@link CanvasState}.
+   */
+  channel: URI;
+  /** Provider-local canvas id this instance was opened from. */
+  canvasId: string;
```
connor4312: Do we need all these ID's? AHP doesn't know what a "provider" is. Channel URI should be enough to uniquely identify an open canvas, as they will already be unique.

types/channels-session/state.ts
```diff
@@ -221,9 +244,182 @@ export interface SessionActiveClient {
...
+export const enum CanvasAvailability {
+  /** The provider is connected and can service requests for this canvas. */
+  Ready = 'ready',
+  /**
+   * The provider is temporarily unavailable (for example, a client provider
+   * that disconnected). The entry is retained so it can be restored when the
+   * provider reconnects; in-flight requests fail until then.
+   */
+  Stale = 'stale',
 }
```
connor4312, line 278: Does this need to be on the session state? If it's only client disconnects, I wonder if we should signal that in general for an `activeClient` vs specifically for the canvas alone

types/channels-session/state.ts
```diff
@@ -221,9 +244,182 @@ export interface SessionActiveClient {
...
+  | CanvasClientProviderSource;
+
+/**
+ * One named action a canvas exposes to the agent, mirroring the shape of a
+ * {@link ToolDefinition} entry. Unique within its owning
+ * `(extensionId, canvasId)`.
+ *
+ * @category Canvas Types
+ */
+export interface SessionCanvasAction {
```
connor4312: "Operation" would match the existing wording we use for changesets

types/channels-canvas/actions.ts
```diff
@@ -0,0 +1,83 @@
...
+ *
+ * @category Canvas Actions
+ * @version 1
+ * @clientDispatchable
+ */
+export interface CanvasMessageAction {
+  type: ActionType.CanvasMessage;
+  /** Opaque, provider-defined message payload. */
+  payload: unknown;
+}
```
connor4312: If this channel encompasses an existing protocol and doesn't really need its own state or OOB messages, we can always do what we did with `mcp://` or `otel://` where we say that it's a channel that exists, refer to its own documentation for the protocol definition -- just with the addition of a `channel: string` for routing data to th eright place.

types/channels-session/state.ts
colbylwilliams: It is consumed — just not by the transport. The agent reads `description` (and the per-operation descriptions) to decide when to open a canvas and what to invoke on it: the host folds it into the canvases section of the agent's system prompt and exposes it through the capability-listing tool, exactly the way `ToolDefinition.description` drives tool selection. The client is the source of truth for what its canvas does, so it publishes the description and the host aggregates it into `SessionState.canvases`. That's why it lives here rather than being server-consumed.

types/channels-session/state.ts
colbylwilliams: Agreed — "extension" is borrowed vocabulary AHP doesn't own. Renamed to `providerId`, documented as an opaque namespace string AHP does not interpret. It's load-bearing rather than removable: `canvasId` is provider-local (two providers can each declare a `diff` canvas), so a namespace qualifier is needed to keep ids unique and to route `canvasOpen`/`canvasInvokeOperation`/`canvasClose` to the right provider. `providerId` lines up with the provider vocabulary already in the surface (`canvasProviders`, `CanvasProviderSource`, `CanvasProviderError`). I also dropped `extensionName` — `displayName` already carries the human-readable label.

types/channels-session/state.ts
colbylwilliams: Agreed. Trimmed `OpenCanvasRef` to `{ channel, canvasId, title?, availability }` — the `channel` URI is now the sole identity (dropped `instanceId`, `providerId`, `extensionName`). I kept `canvasId` because the renderer keys the native surface type off it (editor vs browser vs a custom canvas), so a catalogue view can render the right thing without subscribing to every instance; `title`/`availability` are just for a lightweight list. Full identity now lives only on the `CanvasState` behind the channel.

types/channels-session/state.ts
colbylwilliams: It's not only client disconnects, which is why it's per-instance. A provider can be server-side (host-internal) — no `activeClient` entry at all — yet its canvas can still go `stale` if the host tears the provider down. And one client can provide several canvases with independent lifecycles, so a client-level "connected" bit wouldn't tell a renderer whether the specific instance it's showing is still live. `availability` is the derived per-instance projection of "is this instance's resolved provider reachable", which is what a renderer needs to grey out a stale surface. It's redundant with `activeClient` state for the pure disconnect case, but it's the general signal that also covers the server-provider case — so I'd keep it on the canvas ref/state.

types/channels-session/state.ts
colbylwilliams: Done — renamed to match changesets: `SessionCanvasAction` → `SessionCanvasOperation`, the field `actions` → `operations`, and `canvasInvokeAction` → `canvasInvokeOperation` (`actionName` → `operationName`). I left the two channel state-mutation Actions (`canvas/updated`, `canvas/closeRequested`) as Actions, since those are protocol-level Actions in the same sense changesets keep both an operation vocabulary and channel Actions.

types/channels-canvas/actions.ts
colbylwilliams: Agreed there's no need for its own OOB messages here — but it turns out there's also nothing to encompass. Canvas interaction is just three fixed RPCs (`canvasOpen` / `canvasClose` / `canvasInvokeOperation`) plus availability signals; provider→client never carries opaque data. So instead of a `channel: string` bridge I removed `canvas/message` outright. I also dropped the bespoke `canvasReadResource` path — content now rides the general, symmetrical `resourceRead` (which the relay already carries), so a canvas renders over a relay with no direct host connection. The channel keeps only the two state-mutation Actions (`canvas/updated`, `canvas/closeRequested`).

types/channels-session/state.ts
```diff
@@ -221,9 +244,186 @@ export interface SessionActiveClient {
...
+  inputSchema?: Record<string, unknown>;
+}
+
+/**
+ * One entry in the aggregated {@link SessionState.canvases} registry — a canvas
+ * the agent can open, contributed by a server-side or client-declared provider.
+ *
+ * @category Canvas Types
+ */
+export interface SessionCanvasDeclaration {
```
connor4312, line 353: At this point the canvas is a _lot_ like a `ToolDefinition`. Your comment about the

> the host folds it into the canvases section of the agent's system prompt and exposes it through the capability-listing tool, exactly the way ToolDefinition.description drives tool selection

Makes it sound like it's treated as such. So... should it be? The entrypoint for MCP apps via `_meta.ui`, canvases operating in the same way would make sense to me

types/channels-session/state.ts
```diff
@@ -221,9 +244,186 @@ export interface SessionActiveClient {
...
+  /** Operation name, unique within the owning `(providerId, canvasId)`. */
+  name: string;
+  /** Human-readable description of what the operation does. */
+  description?: string;
+  /**
+   * JSON Schema for the operation's input. Opaque to AHP; mirrors the
+   * {@link ToolDefinition.inputSchema} shape.
+   */
+  inputSchema?: Record<string, unknown>;
+}
```
connor4312, line 345: These also look a lot like tools. In MCP App land, apps can call tools and tools can expose `visibility: ('modell' | 'app')[]` to make them visible to the app. Should the same approach be used here?

types/channels-canvas/state.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Rendering is state-driven: a client renders the canvas by reading
+ * {@link url} and resolving it per the renderer's URL policy — directly for a
+ * reachable address, or over the general `resourceRead` command for a
+ * content-reference address. It never receives a "render this" request.
+ *
+ * @category Canvas State
+ */
+export interface CanvasState {
+  /** Server-assigned instance handle, unique within the session. */
+  instanceId: string;
```
connor4312: Canvases already get a unique URI, what does an additional instanceId add?

types/channels-canvas/state.ts
```diff
@@ -0,0 +1,69 @@
...
+  providerId: string;
+  /** Human-readable canvas name. */
+  displayName?: string;
+  /**
+   * Input the agent supplied when opening the instance. Retained so the
+   * instance can be resumed or rebound after a reconnect.
+   */
+  input?: Record<string, unknown>;
+  /** Current instance title. */
+  title?: string;
```
connor4312: `title` is the general wording we use. I suggest being more explicit like `canvasTitle`/`instanceTitle` over displayName/title

types/channels-canvas/state.ts
```diff
@@ -0,0 +1,69 @@
...
+  /** Provider-defined status string (opaque to AHP). */
+  status?: string;
```
connor4312: should this actually be opaque? Statuses in general usually have different rendering on clients (e.g. normal sessions have active, inputNeeded, errored.)

types/channels-canvas/state.ts
```diff
@@ -0,0 +1,69 @@
...
+   * directly-loadable URL (`https:`, an in-process scheme, `http://localhost`)
+   * or a content-reference URI the renderer resolves with the general
+   * `resourceRead` command. Resolving over `resourceRead` keeps content flowing
+   * entirely over the AHP transport, so a relayed or brokered deployment —
+   * where the host is reachable only over AHP and cannot be dialed directly —
+   * can still serve every byte, with no port or direct connection required. The
+   * renderer dispatches on the scheme and enforces its URL policy. See
+   * {@link /specification/canvas-channel | Canvas Channel}.
+   */
+  url?: string;
```
connor4312: I suggest `contentUri: URI`, any maybe adjusted wording like:

- The URI here must be accessible over resourceRead
- A client may render an `https:` scheme directly, but should fall back to resourceRead if it is inaccessible.

types/channels-session/state.ts
connor4312: Do we need to encode a reason why a provider might be unavailable?

types/channels-session/state.ts
```diff
@@ -221,9 +244,186 @@ export interface SessionActiveClient {
...
+ * `providerId` and {@link SessionCanvasDeclaration.source | `source`} when
+ * folding it into {@link SessionState.canvases}.
+ *
+ * @category Canvas Types
+ */
+export interface ClientCanvasDeclaration {
+  /** Provider-local canvas id, unique within the publishing client. */
+  canvasId: string;
+  /** Human-readable canvas name. */
+  displayName: string;
```
connor4312: We generally have used title here, matching MCP

types/channels-canvas/state.ts
```diff
@@ -0,0 +1,69 @@
...
+export interface CanvasState {
+  /** Server-assigned instance handle, unique within the session. */
+  instanceId: string;
+  /** Provider-local canvas id this instance was opened from. */
+  canvasId: string;
+  /**
+   * Owning provider id — an opaque namespace string AHP does not interpret,
+   * carried so a provider-local {@link canvasId} stays unique across providers.
+   */
+  providerId: string;
```
connor4312, line 43: Isn't the providerId already encoded in the CanvasProviderSource?

types/channels-canvas/commands.ts
```diff
@@ -0,0 +1,149 @@
...
+ * } }
+ * ```
+ */
+export interface CanvasOpenParams extends BaseParams {
+  /** The owning session channel URI (`ahp-session:/<uuid>`). */
+  channel: URI;
+  /** Provider-local canvas id to open. */
+  canvasId: string;
+  /** Owning provider id (opaque to AHP). */
+  providerId: string;
```
connor4312, line 69: Should this be a CanvasProviderSource? Otherwise it's not readily representable for host-owned canvases.

types/channels-canvas/commands.ts
```diff
@@ -0,0 +1,149 @@
...
+ */
+export interface CanvasInvokeOperationParams extends BaseParams {
+  /** The owning session channel URI (`ahp-session:/<uuid>`). */
+  channel: URI;
+  /** Instance handle the operation targets. */
+  instanceId: string;
+  /** Provider-local canvas id of the instance. */
+  canvasId: string;
+  /** Owning provider id (opaque to AHP). */
+  providerId: string;
```
connor4312: Why are these needed here? The URI should be enough right?

types/channels-canvas/commands.ts
```diff
@@ -0,0 +1,149 @@
...
+ * { "jsonrpc": "2.0", "id": 41, "result": {
+ *   "url": "https://acme.example/canvas/inst-7",
+ *   "title": "src/app.ts"
+ * } }
+ * ```
+ */
+export interface CanvasOpenParams extends BaseParams {
+  /** The owning session channel URI (`ahp-session:/<uuid>`). */
+  channel: URI;
+  /** Provider-local canvas id to open. */
```
connor4312, line 66: This should live in session commands.ts if it applies for the session

types/channels-canvas/commands.ts
```diff
@@ -0,0 +1,149 @@
...
+  /** Open input, validated by the provider against its declared schema. */
+  input?: Record<string, unknown>;
+}
+
+/**
+ * Result of the `canvasOpen` command.
+ */
+export interface CanvasOpenResult {
+  /** Initial content address for the instance (see {@link CanvasState.url}). */
+  url?: string;
```
connor4312: Same URI cleanup as in the state here

types/channels-canvas/commands.ts
```diff
@@ -0,0 +1,149 @@
...
+ * @direction Client ↔ Server
+ * @messageType Request
+ * @version 1
+ * @throws `CanvasProviderError` (`-32012`) if the provider has no handler for
+ * the operation or the invocation fails; `data` carries the provider-defined
+ * `{ code, message }`.
+ */
+export interface CanvasInvokeOperationParams extends BaseParams {
+  /** The owning session channel URI (`ahp-session:/<uuid>`). */
+  channel: URI;
```
connor4312: Why session channel vs canvas channel here?

types/channels-canvas/commands.ts
```diff
@@ -0,0 +1,149 @@
...
+ *
+ * @category Commands
+ * @method canvasClose
+ * @direction Client ↔ Server
+ * @messageType Request
+ * @version 1
+ */
+export interface CanvasCloseParams extends BaseParams {
+  /** The owning session channel URI (`ahp-session:/<uuid>`). */
+  channel: URI;
```
connor4312: same

types/channels-canvas/state.ts
colbylwilliams: Nothing — you're right. I dropped `instanceId`; the `ahp-canvas:/<id>` channel URI is now the sole identity, the same way a terminal keys off its own channel URI rather than a separate handle.

types/channels-canvas/commands.ts
colbylwilliams: Agreed. Invoke now takes just the canvas channel URI plus `operationName`/`input` — the host resolves the URI back to the owning provider.

types/channels-canvas/commands.ts
colbylwilliams: It shouldn't be, my bad.  All three (open/invoke/close) now travel on the instance's own `ahp-canvas:/<id>` channel — like `createTerminal` targets the new terminal's channel and `invokeChangesetOperation` targets the changeset's.

types/channels-canvas/commands.ts
colbylwilliams: Same fix — `canvasClose` addresses the canvas channel URI and drops the redundant ids.

types/channels-canvas/commands.ts
colbylwilliams: Done — `CanvasOpenResult.url` → `contentUri: URI` to match the state.

types/channels-session/state.ts
colbylwilliams: Done — renamed to `title` on both `SessionCanvasDeclaration` and `ClientCanvasDeclaration`.

types/channels-canvas/state.ts
colbylwilliams: Renamed `displayName` → `title`. I went with plain `title` to stay consistent with `ToolDefinition.title` and how the field reads elsewhere, rather than `instanceTitle`/`canvasTitle`.

types/channels-canvas/state.ts
colbylwilliams: Adopted `contentUri: URI` and reworked the wording: a directly-loadable address (`https:`, `http://localhost`, an in-process scheme) renders straight into the surface, and a content-reference address resolves over `resourceRead`, which is what carries content over a relay. I stopped short of requiring *every* `contentUri` to be `resourceRead`-accessible with `https` as a fallback — some canvases point at a live address (a localhost preview, an external app) that isn't a `resourceRead` resource, so dispatching on scheme fits the real cases better.

types/channels-canvas/state.ts
colbylwilliams: You're right, carrying is redundant — I dropped `provider: CanvasProviderSource` from `CanvasState`, so the source lives only on the `SessionCanvasDeclaration` now. 

I kept `providerId` because it's a different axis from the source: `clientId` identifies the connection, while `providerId` namespaces a provider *within* a client, since one client can proxy several providers. They coincide for server-side canvases but not client-side ones, so both need to exist.

types/channels-canvas/state.ts
colbylwilliams: `status` is opaque by design — it's provider-defined display text, not a fixed lifecycle the client branches on. The lifecycle-style signal is `availability` (`ready`/`stale`); `status` is free-form text the provider owns, so a fixed enum would fight the cases it's meant for (a provider's own progress or among-many labels). Session `active`/`inputNeeded`/`errored` drives host behavior, which this deliberately doesn't.

types/channels-session/state.ts
colbylwilliams: `ready`/`stale` are the two a renderer acts on (show it, or show it dimmed). A reason would be presentational only. There's no consumer yet and it's easy to add `reason?` later if one shows up.  Happy to add now if you feel strongly.

types/channels-session/state.ts
colbylwilliams: Not exactly - a canvas is a durable, subscribable resource with its own channel, state, and lifecycle — closer to a terminal or changeset than to a one-shot tool result — and AHP already models the MCP-app UI hint separately (`ClientCapabilities.mcpApps`, `McpServerCustomization.mcpApp`). As a data model I think the two should stay distinct rather than folding canvases into `ToolDefinition`/`_meta.ui`. If the broader plan is to route canvas discovery through the same entrypoint, that's a direction call I'll happily follow your lead on.

types/channels-session/state.ts
colbylwilliams: The `visibility: ('model'|'app')[]` flag gates whether the model or the app UI may call an MCP-app *tool*. Canvas operations are always agent-invoked, so there's no second caller to gate — the flag would carry a single value today. If we later let the rendered surface invoke operations itself, that's when `visibility` would earn its place.

types/channels-canvas/commands.ts
colbylwilliams: A host-owned canvas is already representable — the host declares its own canvases with a `providerId`, their `source` being `{ kind: 'server' }` on the declaration. `canvasOpen` only needs to name which provider + canvas to open, and `(providerId, canvasId)` covers host- and client-owned alike; the host resolves the source from the declaration.

Passing `CanvasProviderSource` here would re-introduce the `clientId` the channel URI now makes redundant, so `(providerId, canvasId)` is the tighter shape.

types/channels-canvas/commands.ts
colbylwilliams: I kept the three methods in `channels-canvas/commands.ts`, next to the channel they operate on — mirroring `createTerminal` in `channels-terminal/commands.ts`, which likewise brings its channel into being. 

Now that they target the `ahp-canvas:/<id>` channel rather than the session channel, grouping them with the canvas channel is the consistent spot — but if you'd rather group by initiating scope, I'll move them.

types/channels-canvas/state.ts
connor4312: Is this more akin to our existing `activity?: string` than status?

types/channels-session/state.ts
```diff
@@ -221,9 +244,185 @@ export interface SessionActiveClient {
...
+ *
+ * @category Canvas Types
+ */
+export interface SessionCanvasDeclaration {
+  /**
+   * Owning provider id — an opaque namespace string AHP does not interpret,
+   * carried so a provider-local {@link canvasId} stays unique across providers.
+   * Stable across declarations and instances.
+   */
+  providerId: string;
```
connor4312, line 359: I still don't understand what the providerId is. It ClientCanvasDeclaration says "The host derives the `providerId` and {@link SessionCanvasDeclaration.source | `source`}" and "Canvas declarations this client contributes as a provider" which makes me think the providerId is the ID of the ActiveClient that was providing this canvas, but then we use this separately from the CanvasProviderSource which actually _does_ contain the client ID.

types/channels-session/state.ts
```diff
@@ -221,9 +244,185 @@ export interface SessionActiveClient {
...
+ * The lighter declaration shape a client publishes on
+ * {@link SessionActiveClient.canvasProviders}. The host derives the
+ * `providerId` and {@link SessionCanvasDeclaration.source | `source`} when
+ * folding it into {@link SessionState.canvases}.
+ *
+ * @category Canvas Types
+ */
+export interface ClientCanvasDeclaration {
+  /** Provider-local canvas id, unique within the publishing client. */
+  canvasId: string;
```
connor4312, line 387: on ^ saying the "provider local ID" in the context of just the IDs given from the client make me think that when we say provider we mean client.

types/channels-canvas/state.ts
colbylwilliams: Yeah, you're right. Changed.

types/channels-session/state.ts
joshmouch: I went spelunking through these types and I think you've nailed it - the canvas declaration really is a tool, and the PR says so itself: `SessionCanvasOperation`'s header literally reads "mirroring the shape of a `{@link ToolDefinition}` entry," and both `inputSchema` fields echo it. So yeah - route the declaration and its operations through the tool entrypoint; the types are basically begging for it.

One catch before you commit to that: the mirror isn't actually built yet. Canvas `inputSchema` is `Record<string, unknown>`, while `ToolDefinition.inputSchema` is `{ type: 'object'; properties?: Record<string, object>; required?: string[] }` - so a `SessionCanvasOperation` won't assign to a `ToolDefinition` today. The doc's writing a check the types don't cash. If you go the tool route, tighten that schema into a real `ToolDefinition`, or the mirror just lives in the comments.

Where I'd stop the analogy is the open instance - and this one isn't taste, it's the type system drawing the line for us. `OpenCanvasRef` is `{ channel: URI; canvasId; title?; availability }` with `availability: 'ready' | 'stale'`, and `CanvasState` sits in the `Snapshot.state` union in `types/common/state.ts` (which this PR touches). That union membership is what lets a channel replay on reconnect. `ToolCallState` isn't in there - it has no channel of its own. It lives inside `ChatState` as a `ToolCallResponsePart`, and the one spot the session surfaces a *running* tool call, it copies it onto a session input-request (`SessionToolClientExecutionRequest.toolCall` - which exists precisely so a client can pick up the work "without subscribing to the owning chat"). There's no `ahp-toolcall:/<id>` to subscribe to; there's an `ahp-canvas:/<id>`. A channel URI and a `stale` bit just have nowhere to live on a `ToolDefinition`.

You drew this exact line yourself in #336, for what it's worth: `McpAuthRequirement` is one OAuth challenge from two vantage points, and `McpServerAuthRequiredState` / `ToolCallAuthRequiredState` stay separate on purpose - "dispatched independently and MAY be true at the same time, or not." Same story here.

So, my two cents: unify the *declaration* with the tool entrypoint (and make its schema a real `ToolDefinition` while you're in there), but let the open *instance* keep its own channel. The tool shape is perfect for discovery and wrong for lifetime.

PR #299, opened 2026-07-02
Changeset - add optional `contentSha256` to `ContentRef`

lszomoru:


types/common/state.ts
```diff
@@ -252,6 +252,8 @@ export interface ContentRef {
   sizeHint?: number;
   /** Content MIME type */
   contentType?: string;
+  /** Content SHA-256 hash */
+  contentSha256?: string;
```
connor4312: I think we should name this `nonce?: string` and let the implementor decide. This is the same thing done in `ClientPluginCustomization`

connor4312 (approved):

aeschli (approved):

PR #305, opened 2026-07-07
fix: make generated JSON Schemas self-contained and strict-oneOf-safe

### LINKED ISSUE #302: JSON Schema generator emits invalid schemas: spurious `{}` oneOf branch for union aliases + dangling `$ref`s in notifications.schema.json
Two related, pre-existing correctness problems in the JSON Schema generator (`scripts/generate-json-schema.ts`). Both are repo-wide (not specific to any one type) and were surfaced by review on #298. The generated schemas are reference/documentation artifacts today — there is no CI gate that resolves `$ref`s or runs strict `oneOf` validation, which is why these have gone unnoticed — but any consumer that validates payloads against them will hit both.

## 1. Spurious empty `{}` branch in `oneOf` for union type aliases

Every union *type alias* is emitted with a leading empty-schema branch, e.g.:

```json
"CanvasProviderSource": {
  "oneOf": [
    {},
    { "$ref": "#/$defs/CanvasServerProviderSource" },
    { "$ref": "#/$defs/CanvasClientProviderSource" }
  ]
}
```

Because `{}` matches *any* value, every instance matches ≥2 `oneOf` branches, so strict `oneOf` validation always fails.

**Root cause:** a union type alias's `getTypeNode().getText()` includes the leading `|` (i.e. `| A\n  | B`). `splitUnionType` then produces an empty-string first element, and `typeTextToSchema('')` falls through to `return {}`. The union handler filters out `'undefined'` parts but not empty-string parts.

**Scope:** `state.schema.json` (15 `$defs` — 14 on `main` before the canvas addition) and `actions.schema.json` (17 `$defs`). Affected unions include `ToolCallState`, `ToolCallConfirmationState`, `SessionInputRequest`, `JsonPrimitive`, `ChatOrigin`, `MessageAttachment`, `ResponsePart`, `ToolResultContent`, `Customization`, `McpServerState`, `TerminalContentPart`, `CanvasProviderSource`, …

**Likely fix:** drop empty parts in the union branch, e.g. `parts.filter(p => p !== 'undefined' && p !== '')` (or strip a leading `|` before splitting).

## 2. Dangling `$ref`s in `notifications.schema.json`

`notifications.schema.json` references ~95 `$defs` it never defines (e.g. `ToolCallState`, `SessionInputRequest`, `MessageAttachment`, `URI`, `JsonPrimitive`, `ChatOrigin`, every `*Kind`/status enum member, and now `CanvasProviderSource`), leaving unresolved `$ref`s and making the file non-self-contained / invalid for standalone validation.

**Root cause:** the notifications schema generator only inlines a subset of `$defs` (the interfaces it directly walks) and does not transitively pull in referenced type aliases/enums the way the state/actions generators do.

**Likely fix:** transitively collect and inline every referenced `$def`, or `$ref` them into a shared/base schema document.

## Note

Discovered while reviewing #298 (Canvas channel). The canvas types (`CanvasProviderSource`, `CanvasState`) reproduce the identical pre-existing pattern, so they are intentionally left following the established convention in that PR and tracked here for a focused, generator-wide fix.

joshmouch:
Fixes #302.

Both bugs are fixed generator-wide, plus one scope correction and a latent crash surfaced along the way.

## Bug 1: empty `{}` oneOf branch

A union type alias's printed type text carries a leading `|` (`type X =\n  | A\n  | B`), so `splitUnionType` yielded an empty first element that fell through `typeTextToSchema` to `{}`. Every union alias's `oneOf` therefore led with a matches-anything branch, so every instance matched two or more branches and strict `oneOf` validation always failed.

While fixing this I hit a pre-existing infinite recursion the union path could already trigger: an inline object with a nested union field (e.g. `side?: 'before' | 'after'`) contains a `|`, so it entered the union branch; `splitUnionType` correctly kept it as a single part (the `|` is nested inside `{}`), and the single-part early-return re-processed the identical string forever. The union branch now fires only for a genuine top-level union (two or more parts) and otherwise falls through to the inline-object / reference handling.

## Bug 2: dangling `$ref`s (broader than the issue scopes it)

One correction on scope: the dangling-ref problem is not unique to `notifications.schema.json`. Every generated file has them. Counts before this PR:

| file | dangling `$ref`s |
|---|---|
| state.schema.json | 74 |
| actions.schema.json | 154 |
| commands.schema.json | 182 |
| notifications.schema.json | 95 |
| errors.schema.json | 103 |
| **total** | **608** |

And most of those (roughly 60-75% per file) are a second root cause the issue does not mention: the generator never walked enums. Any property typed as a bare `const enum` (`SessionStatus`, `TurnState`, ...) or an enum-member-literal discriminant (`type: ActionType.AnnotationsEntryRemoved`) hit the capitalized-identifier fallback in `typeTextToSchema` and was emitted as a `$ref` to a `$def` that was never defined, in any file. The remaining dangling refs were type aliases (`URI`, `JsonPrimitive`, ...) that only two of the five generators transitively collected.

The fix has three parts:
- Enum-member accesses inline as `{ "const": <member value> }` (so `type: ActionType.Foo` becomes `{ "const": "root/agentsChanged" }` rather than a dangling ref).
- A fixpoint `resolveMissingDefs` pass runs after each schema is built and backfills every referenced `$def` that is not yet present, resolving interfaces, type aliases (including `URI = string`), and whole enums (`SessionStatus` becomes `{ "type": "number", "enum": [1, 2, 8, 24, 32, 64] }`, with its bitwise member values evaluated). It throws loudly if a referenced name resolves to none of those, so a future gap fails the build instead of silently emitting a dangling ref.
- `null` union members now render as `{ "type": "null" }` instead of `{}`.

I chose transitive inline collection (option A from the issue) rather than cross-file `$ref`s (option B), since every existing schema file is already self-contained and I did not want to introduce a new cross-document architecture.

## Result

After this PR, across all five files: **0 dangling `$ref`s** (was 608) and **0 empty `{}` oneOf branches** (was 30).

There was no test coverage for the schema generator and no CI gate that would have caught any of this, so I added `scripts/generate-json-schema.test.ts` (wired into `npm test`) asserting that no generated schema has an empty `{}` oneOf branch or a dangling `$ref`. It is dependency-free (no new packages) and fails if either bug is reintroduced.

`npm test` passes (290 tests, typecheck, lint, changelog + release-metadata gates). All five `schema/*.json` files are regenerated in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,15 @@ Spec version: `0.5.2`
   `id` and sets that entry's `enabled`; the effective state of a child is
   `container.enabled && (child.enabled ?? true)`.
 
+### Fixed
+
+- The generated JSON Schema artifacts (`schema/*.json`) are now self-contained
+  and strict-`oneOf`-safe. The generator no longer emits an empty `{}` branch in
+  a union type alias's `oneOf` (which matched any value and defeated strict
+  `oneOf` validation), backfills the `enum`, enum-member, and type-alias `$ref`s
+  that were previously dangling across all five schema files, and renders `null`
+  members as `{ "type": "null" }` (#302).
+
 ## [0.5.1] — 2026-07-02
 
 Spec version: `0.5.1`
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "docs:preview": "vitepress preview docs",
     "typecheck": "tsc --noEmit -p types/tsconfig.json",
     "lint": "eslint",
-    "test": "npm run typecheck && npm run lint && npm run verify:release-metadata && npm run verify:changelog && npx c8 --include types/reducers.ts --check-coverage --branches 100 tsx --test types/*.test.ts types/version/*.test.ts"
+    "test": "npm run typecheck && npm run lint && npm run verify:release-metadata && npm run verify:changelog && npx c8 --include types/reducers.ts --check-coverage --branches 100 tsx --test types/*.test.ts types/version/*.test.ts scripts/*.test.ts"
   },
   "repository": {
     "type": "git",
--- a/schema/actions.schema.json
+++ b/schema/actions.schema.json
@@ -54,7 +54,7 @@
       "description": "Fired when available agent backends or their models change.",
       "properties": {
         "type": {
-          "$ref": "#/$defs/ActionType.RootAgentsChanged"
+          "const": "root/agentsChanged"
         },
         "agents": {
           "type": "array",
@@ -74,7 +74,7 @@
       "description": "Fired when the number of active sessions changes.",
       "properties": {
         "type": {
-          "$ref": "#/$defs/ActionType.RootActiveSessionsChanged"
+          "const": "root/activeSessionsChanged"
         },
         "activeSessions": {
           "type": "number",
@@ -91,7 +91,7 @@
       "description": "Fired when the list of known terminals changes.\n\nFull-replacement semantics: the `terminals` array replaces the previous\n`terminals` entirely.",
       "properties": {
         "type": {
-          "$ref": "#/$defs/ActionType.RootTerminalsChanged"
+          "const": "root/terminalsChanged"
         },
         "terminals": {
           "type": "array",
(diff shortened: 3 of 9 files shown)


connor4312 (approved):

vritant24 (approved):

roblourens (approved):

connor4312:
thanks!

PR #308, opened 2026-07-07
Add optional `_meta` slot to `SystemNotificationResponsePart`

### LINKED ISSUE #307: Add a `_meta` slot to `SystemNotificationResponsePart` for machine-readable trigger metadata
### Summary

`SystemNotificationResponsePart` conveys a system notification as human-readable `content` only. There is no machine-readable indication of *what triggered* it, so a client cannot reliably categorize, icon, group, filter, or localize these parts without parsing natural-language text. Adding a `_meta` slot — consistent with the `_meta` convention already used across AHP state — closes the gap.

### Current state

`SystemNotificationResponsePart` (`types/channels-chat/state.ts`) is:

```ts
export interface SystemNotificationResponsePart {
  kind: ResponsePartKind.SystemNotification;
  content: StringOrMarkdown;
}
```

The type's own doc gives examples that are inherently categorical:

> Examples include "background subagent X completed" or "task Y was cancelled".

But "subagent completed" versus "task cancelled" is expressed only inside the prose `content`. The discriminant `kind` says merely "this is a system notification"; it does not say which *kind of* event produced it.

### The gap

A client that wants to treat these parts as more than opaque text — a distinct icon, severity, grouping, collapse/dismiss behavior, or localized rendering — has to string-match `content`, which is brittle and defeats localization.

`_meta` is the established escape hatch on AHP *state* carriers: `Message`, `ChatState`, `SessionState`, `SessionSummary`, `SessionModelInfo`, `RootState`, and `ToolDefinition` all carry `_meta?: Record<string, unknown>` (documented as mirroring the MCP `_meta` convention). No response-part type carries one today — and `SystemNotificationResponsePart`, whose entire purpose is to surface an out-of-band harness event, is the part that most needs it.

### Proposal

Add the standard `_meta` slot:

```ts
export interface SystemNotificationResponsePart {
  kind: ResponsePartKind.SystemNotification;
  content: StringOrMarkdown;
  /**
   * Reserved for implementation-specific metadata, following the MCP `_meta`
   * convention. A host MAY attach a machine-readable descriptor of what
   * triggered the notification so clients can render it richly without
   * parsing `content`.
   */
  _meta?: Record<string, unknown>;
}
```

This is the minimal, convention-consistent change and unblocks hosts immediately — each namespaces its trigger descriptor under its own `_meta` key.

### Optional follow-on: a portable trigger discriminator

If there is cross-host appetite, a small optional typed field for the common, portable categories the doc already names could be standardized later:

```ts
// Sketch only — not required for the _meta ask above.
interface SystemNotificationTrigger {
  category: 'subagentCompleted' | 'taskCancelled' | 'backgroundTaskFinished' | string;
  // …category-specific fields…
}
```

The trigger vocabulary is open-ended and partly harness-specific, so `_meta` is the right first step; a typed field can follow if a portable subset proves worthwhile. Deferring to maintainers on whether to standardize any categories.

### Compatibility

Purely additive and optional. Existing clients ignore `_meta`; servers that omit it stay compliant.

Happy to open a PR for the `_meta` addition.

colbylwilliams:
## Summary

Closes #307.

`SystemNotificationResponsePart` conveys a harness event as human-readable `content` only. There is no machine-readable indication of *what triggered* it, so a client cannot categorize, icon, group, filter, or localize these parts without string-matching natural-language prose. This adds the standard `_meta` escape hatch to close that gap.

```ts
export interface SystemNotificationResponsePart {
  kind: ResponsePartKind.SystemNotification;
  content: StringOrMarkdown;
  /**
   * Additional provider-specific metadata for this notification.
   *
   * A host MAY attach a machine-readable descriptor of what triggered the
   * notification so clients can categorize, icon, group, filter, or localize
   * it without parsing `content`. Clients MAY look for well-known keys here to
   * provide enhanced UI, and MUST render coherently from `content` alone when
   * `_meta` is absent or unrecognized.
   */
  _meta?: Record<string, unknown>;
}
```

## Why

- `_meta?: Record<string, unknown>` is the established escape hatch across AHP state carriers (`Message`, `ChatState`, `UsageInfo`, `ErrorInfo`, `RootState`, `ToolDefinition`, session types, …), mirroring the MCP `_meta` convention. No response-part type carried one before — and the system notification, whose whole purpose is to surface an out-of-band harness event, is the part that most needs it.
- It fits the doctrine's "escape hatches are explicit" principle: provider-specific metadata is allowed but never required for the baseline experience. A minimal client ignores `_meta` and still renders coherently from `content`; a typed trigger discriminator can graduate out of `_meta` later if a portable subset proves worthwhile.

## Compatibility

Purely additive and optional. Existing clients ignore `_meta`; hosts that omit it stay compliant. No reducer change is required — `SystemNotificationResponsePart` is delivered via the existing `ChatResponsePartAction`, whose reducer appends the part wholesale, so `_meta` rides through unchanged.

## Changes

- **`types/channels-chat/state.ts`** — add the `_meta` slot with a convention-matching doc comment.
- **Generated mirrors** — regenerated the JSON Schema and all five client mirrors (Rust, Kotlin, Swift, TypeScript, Go).
- **`docs/guide/state-model.md`** — document `SystemNotificationResponsePart` in the Response Parts section (it was previously missing) and describe the `_meta` slot.
- **CHANGELOGs** — added an `### Added` entry under `## [0.5.2] — Unreleased` in the spec changelog and every per-client changelog.

### DIFF
--- a/clients/go/ahptypes/state.generated.go
+++ b/clients/go/ahptypes/state.generated.go
@@ -1645,6 +1645,14 @@ type SystemNotificationResponsePart struct {
 	Kind ResponsePartKind `json:"kind"`
 	// The text of the system notification
 	Content StringOrMarkdown `json:"content"`
+	// Additional provider-specific metadata for this notification.
+	//
+	// A host MAY attach a machine-readable descriptor of what triggered the
+	// notification so clients can categorize, icon, group, filter, or localize
+	// it without parsing `content`. Clients MAY look for well-known keys here to
+	// provide enhanced UI, and MUST render coherently from `content` alone when
+	// `_meta` is absent or unrecognized.
+	Meta map[string]json.RawMessage `json:"_meta,omitempty"`
 }
 
 // A resolved input request (elicitation) recorded in the turn transcript.
--- a/clients/kotlin/src/main/kotlin/com/microsoft/agenthostprotocol/generated/State.generated.kt
+++ b/clients/kotlin/src/main/kotlin/com/microsoft/agenthostprotocol/generated/State.generated.kt
@@ -2289,7 +2289,18 @@ data class SystemNotificationResponsePart(
     /**
      * The text of the system notification
      */
-    val content: StringOrMarkdown
+    val content: StringOrMarkdown,
+    /**
+     * Additional provider-specific metadata for this notification.
+     *
+     * A host MAY attach a machine-readable descriptor of what triggered the
+     * notification so clients can categorize, icon, group, filter, or localize
+     * it without parsing `content`. Clients MAY look for well-known keys here to
+     * provide enhanced UI, and MUST render coherently from `content` alone when
+     * `_meta` is absent or unrecognized.
+     */
+    @SerialName("_meta")
+    val meta: Map<String, JsonElement>? = null
 )
 
 @Serializable
--- a/clients/rust/crates/ahp-types/src/state.rs
+++ b/clients/rust/crates/ahp-types/src/state.rs
@@ -2062,6 +2062,15 @@ pub struct ReasoningResponsePart {
 pub struct SystemNotificationResponsePart {
     /// The text of the system notification
     pub content: StringOrMarkdown,
+    /// Additional provider-specific metadata for this notification.
+    ///
+    /// A host MAY attach a machine-readable descriptor of what triggered the
+    /// notification so clients can categorize, icon, group, filter, or localize
+    /// it without parsing `content`. Clients MAY look for well-known keys here to
+    /// provide enhanced UI, and MUST render coherently from `content` alone when
+    /// `_meta` is absent or unrecognized.
+    #[serde(rename = "_meta", default, skip_serializing_if = "Option::is_none")]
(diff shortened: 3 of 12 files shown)


roblourens (dismissed):
This is fine, but is there some `origin` that would make sense as a real property?

connor4312:
> There is no machine-readable indication of what triggered it, so a client cannot categorize, icon, group, filter, or localize these parts without string-matching natural-language prose

Yea, there is still not, at least not in a standard way. Do we want to actually be more descriptive here?

colbylwilliams:
@roblourens @connor4312

Agreed.  I started with `_meta` here, but the [linked issue](https://github.com/microsoft/agent-host-protocol/issues/307) calls out `kind` or something similar (like `origin`) as a proposed follow up.  Happy to do both here, or separately.  Just say the word.

connor4312 (dismissed):

connor4312:
Let's start here but plan to formalize more data in the future

connor4312 (dismissed):

roblourens (approved):

PR #314, opened 2026-07-07
Add shell exit tool result metadata

anthonykim1:
Upstream changes from: https://github.com/microsoft/vscode/pull/323922
Also seems like we were behind on `npm run generate` for changeset related PRs?


types/channels-chat/state.ts
```diff
@@ -1240,6 +1241,25 @@ export interface ToolResultTerminalContent {
...
+export interface ToolResultShellExitContent {
+  type: ToolResultContentType.ShellExit;
+  /** Shell id, as assigned by the runtime */
+  shellId: string;
+  /** Exit code from the completed shell command */
+  exitCode: number;
+  /** Working directory where the shell command was executed */
+  cwd?: string;
+  /** Output preview associated with the shell command, if available */
+  outputPreview?: string;
```
roblourens: Should we separate this so that terminal output can be streamed?

types/channels-chat/state.ts
```diff
@@ -1240,6 +1241,25 @@ export interface ToolResultTerminalContent {
...
 
+/**
+ * Shell command exit metadata emitted by a shell tool.
+ *
+ * @category Tool Result Content
+ */
+export interface ToolResultShellExitContent {
+  type: ToolResultContentType.ShellExit;
+  /** Shell id, as assigned by the runtime */
+  shellId: string;
```
roblourens: Why is this required? Do we display it?

types/channels-chat/state.ts
```diff
@@ -1240,6 +1241,25 @@ export interface ToolResultTerminalContent {
...
+  /** Exit code from the completed shell command */
+  exitCode: number;
```
roblourens: Might make sense for this to be optional too

types/channels-chat/state.ts
```diff
@@ -1240,6 +1241,25 @@ export interface ToolResultTerminalContent {
...
+  /** Working directory where the shell command was executed */
+  cwd?: string;
```
roblourens: Use the URI type

roblourens (dismissed):

types/channels-chat/state.ts
```diff
@@ -1173,6 +1173,7 @@ export const enum ToolResultContentType {
   Resource = 'resource',
   FileEdit = 'fileEdit',
   Terminal = 'terminal',
+  ShellExit = 'shell_exit',
```
roblourens: I'd suggest camelCase to match the other ones

anthonykim1: Changed them here: https://github.com/microsoft/agent-host-protocol/pull/314/commits/31139a6928fc8e411c26e6859cd568e992dc6bf7

types/channels-chat/state.ts
anthonykim1: cwd now uses URI type: https://github.com/microsoft/agent-host-protocol/pull/314/commits/31139a6928fc8e411c26e6859cd568e992dc6bf7

types/channels-chat/state.ts
anthonykim1: exit code is now optional: https://github.com/microsoft/agent-host-protocol/pull/314/commits/31139a6928fc8e411c26e6859cd568e992dc6bf7

types/channels-chat/state.ts
anthonykim1: We probably don't need it. We currently dont use it for displaying. The <shell id ___ exited with ____> message you would see would probably go away once we start using `shell_exit.outputPreview` instead of `ToolResultContentType.Text` used to display output for inline terminal. 

Maybe we can add them back in future if we find use cases for copilot or other harness? 

Removed it as part of: https://github.com/microsoft/agent-host-protocol/pull/314/commits/31139a6928fc8e411c26e6859cd568e992dc6bf7

types/channels-chat/state.ts
anthonykim1: Not sure if we should take streaming as part of `ToolResultShellExitContent` in the first place🧐
Maybe under something new like `ToolResultShellProgressContent`?

I was thinking`ToolResultShellExitContent` should ideally be for when shell comes back after some kind of exit/completion.

types/channels-chat/state.ts
```diff
@@ -1240,6 +1241,23 @@ export interface ToolResultTerminalContent {
...
+  type: ToolResultContentType.ShellExit;
+  /** Exit code from the completed shell command, if reported by the runtime */
+  exitCode?: number;
+  /** Working directory where the shell command was executed */
+  cwd?: URI;
+  /** Output preview associated with the shell command, if available */
+  outputPreview?: string;
+  /** Whether `outputPreview` is known to be incomplete or truncated */
+  outputTruncated?: boolean;
+}
```
connor4312: So this is essentially a 'historic' view of what was a terminal result? Maybe we model it like `ToolResultTerminalCompleteContent` and optionally let it reference a terminal `URI` so that consumers can use it to reference a previously-running terminal, if any.

I would probably do `preview`/`truncated` and drop the `output` -- it's already a tool result

anthonykim1: Little hesitant to use the word 'terminal' here.
Copilot CLI, and other CLIs (codex, cc) have moved away from using pty on behalf of direct shell child processes to execute their commands. 

So they're not really keeping a interactive terminal with persistent state (hence the naming `shell` instead of `terminal`)
We could go for `ToolResultShellCompletionContent`, but I thought exit was more granular. (process exiting for <x> reason vs. complete). Curious of your thought.

Perhaps it's about good time to re-visit `Terminal` prefixed interfaces 🧐
* Wondering if using pty vs. spawning child processes is some implementation detail for all these "shell tools" that AHP should just abstract under terminal.

Adjusted the naming for for preview/truncated: [36bb188](https://github.com/microsoft/agent-host-protocol/pull/314/commits/36bb188b546faa9731cb48d42f232761dabd8204)

anthonykim1: Made further change on [f4d684a](https://github.com/microsoft/agent-host-protocol/pull/314/commits/f4d684a19945c7db4d8a83e7de2418d152372d54) and switched to `ToolResultTerminalCompleteContent`

types/channels-chat/state.ts
```diff
@@ -1230,6 +1231,9 @@ export interface ToolResultFileEditContent extends FileEdit {
  * Clients can subscribe to the terminal's URI to stream its output in real
  * time, providing live feedback while a tool is executing.
  *
+ * The referenced channel is not necessarily PTY-backed; clients MUST NOT
+ * assume input, resize, or VT rendering semantics are available.
+ *
```
connor4312: Let's not cross this bridge yet

types/channels-chat/state.ts
```diff
@@ -1240,6 +1244,40 @@ export interface ToolResultTerminalContent {
...
+ * @category Tool Result Content
+ */
+export interface ToolResultTerminalCompleteContent {
+  type: ToolResultContentType.TerminalComplete;
+  /**
+   * Terminal channel URI that carried live output for this command, if one
+   * was exposed. Clients that subscribe to this channel SHOULD NOT assume a
+   * PTY-backed terminal with input or resize support.
+   */
```
connor4312, line 1263: Let's just spec this as an ahp-terminal command for now

connor4312 (approved):

types/channels-chat/state.ts
anthonykim1: future issue to fully address: https://github.com/microsoft/agent-host-protocol/issues/322

types/channels-chat/state.ts
anthonykim1: future issue to fully address: https://github.com/microsoft/agent-host-protocol/issues/322

Issue #316, opened 2026-07-08
Hoist `_meta` onto `CustomizationBase` so every customization type carries provider metadata uniformly


colbylwilliams:
### Summary

`_meta` exists on exactly one customization type. `AgentCustomization` has a `_meta` slot ("Additional provider-specific metadata for this custom agent. Mirrors the MCP `_meta` convention"); no other customization type does. So a host with provider-specific metadata for a plugin, directory, skill, rule, hook, prompt, or MCP server has no escape hatch, while an agent does — an asymmetry with no obvious justification.

This proposes hoisting `_meta` onto the shared `CustomizationBase` so every customization type carries one uniformly, and dropping the now-redundant per-type declaration on `AgentCustomization`.

### Current state in the spec

`types/channels-session/state.ts`:

```ts
interface CustomizationBase {
  id: string;
  uri: URI;
  name: string;
  icons?: Icon[];
  range?: TextRange;
  // no _meta
}

export interface AgentCustomization extends ChildCustomizationBase {
  type: CustomizationType.Agent;
  description?: string;
  model?: string;
  tools?: string[];
  disableModelInvocation?: boolean;
  disableUserInvocation?: boolean;
  _meta?: Record<string, unknown>; // the ONLY customization type with _meta
}
```

Every other customization — `PluginCustomization`, `DirectoryCustomization`, `SkillCustomization`, `PromptCustomization`, `RuleCustomization`, `HookCustomization`, `McpServerCustomization`, and `ClientPluginCustomization` — extends a base that has no `_meta`, so none of them can carry provider-specific metadata.

### Proposal

Move `_meta` from `AgentCustomization` up to `CustomizationBase`, so all customization types inherit it:

```ts
interface CustomizationBase {
  id: string;
  uri: URI;
  name: string;
  icons?: Icon[];
  range?: TextRange;
  /**
   * Additional provider-specific metadata for this customization.
   * Mirrors the MCP `_meta` convention. Optional and opaque to the
   * protocol; producers and consumers agree on its contents
   * out-of-band.
   */
  _meta?: Record<string, unknown>;
}

export interface AgentCustomization extends ChildCustomizationBase {
  type: CustomizationType.Agent;
  description?: string;
  model?: string;
  tools?: string[];
  disableModelInvocation?: boolean;
  disableUserInvocation?: boolean;
  // _meta now inherited from CustomizationBase
}
```

[...]

Issue #303, opened 2026-07-03
Add a native `argumentHint` field to `SkillCustomization` for slash-command argument help


colbylwilliams:
## Summary

`SkillCustomization` describes a host-discovered skill so a client can list it, show its `description`, and (via `disableModelInvocation`) reason about model auto-invocation. What it can't express is the **argument hint** for a user-invocable, slash-command-style skill — the short "how do I call this" string that a command palette renders next to the command (e.g. `/deploy <env> [--dry-run]`).

There's no field for it and no generic escape hatch: unlike `AgentCustomization`, `SkillCustomization` has no `_meta` slot, so a host that receives an argument hint from its runtime has to drop it. A client building a slash-command palette then can't show the user what arguments a command expects.

## Current state in the spec

- `SkillCustomization` carries `id` / `uri` / `name` / `icons?` / `range?` / `description?` / `disableModelInvocation?`.
- `RuleCustomization` already shows the spec's established style of modeling skill/rule behavior with first-class, type-specific fields (`alwaysApply`, `globs`) rather than a generic bag.
- `AgentCustomization` has a `_meta` slot; `SkillCustomization` does not — so there isn't even a non-native fallback for skill-level metadata.

## Proposal

Add an optional, display-only field to `SkillCustomization`:

```ts
export interface SkillCustomization extends CustomizationBase {
  type: CustomizationType.Skill;
  description?: string;
  disableModelInvocation?: boolean;
  /**
   * Freeform hint describing the arguments a user-invocable skill
   * expects, for display in a slash-command palette or help text
   * (e.g. "<env> [--dry-run]"). Absent when the skill takes no
   * arguments or the source declares no hint. Display-only: hosts
   * neither parse nor enforce it.
   */
  argumentHint?: string;
}
```

- **Optional**, so existing producers/consumers are unaffected until they opt in.
- **Display-only**: a hint for humans, not a parseable arg schema. Hosts don't validate against it.
- Absent ⇒ no hint (skill takes no arguments, or the source declared none).

## Concrete grounding (public, vendor SDK example)

A representative agent SDK already reports this per skill: `session.skills_loaded` skill entries carry an `argumentHint` string sourced from the skill's `argument-hint` frontmatter, right alongside `description`, `enabled`, and `userInvocable`. So the data exists at the source; only the AHP carrier field is missing.

## Relationship to #285

[...]

PR #317, opened 2026-07-08
Add native `version` field to `PluginCustomization`

### LINKED ISSUE #315: Add native `version` field to `PluginCustomization`
### Summary

`PluginCustomization` mirrors an [Open Plugins](https://open-plugins.com/) package and carries the plugin's `name`, but not its `version` — even though the Open Plugins manifest (`plugin.json`) has a first-class `version` field. So a host that knows the version of a plugin it resolved has nowhere to put it, and the value is dropped on the wire.

This proposes adding a native, optional `version` field to `PluginCustomization`.

### Current state in the spec

`types/channels-session/state.ts`:

```ts
interface CustomizationBase {
  id: string;
  uri: URI;
  name: string;
  icons?: Icon[];
  range?: TextRange;
}

interface ContainerCustomizationBase extends CustomizationBase {
  enabled: boolean;
  clientId?: string;
  load?: CustomizationLoadState;
  children?: ChildCustomization[];
}

export interface PluginCustomization extends ContainerCustomizationBase {
  type: CustomizationType.Plugin;
  // carries name (via CustomizationBase) but no version
}
```

### Proposal

A plugin version is a first-class attribute of a plugin, not incidental metadata — the Open Plugins manifest models it natively (`"version": "1.2.0"`, semver, "used by tools to determine update availability"), right next to `name`. AHP's `PluginCustomization` already mirrors that manifest and already carries `name`; `version` is the missing sibling. It earns a native field rather than a `_meta` entry, for the same reason `RuleCustomization` gets `alwaysApply`/`globs` and `AgentCustomization` gets `model`/`tools` instead of a generic bag.

```ts
export interface PluginCustomization extends ContainerCustomizationBase {
  type: CustomizationType.Plugin;
  /**
   * Version of the plugin, sourced from the Open Plugins manifest's
   * `version` field (semver). Absent when the manifest declares no
   * version (the field is optional) or the source has no version
   * concept. Provenance / display only — the host does not parse or
   * enforce it.
   */
  version?: string;
}
```

Because `ClientPluginCustomization extends PluginCustomization`, a client publishing a plugin can carry `version` too; the host populates it on the resolved plugin that appears in `SessionState.customizations`.

### Concrete grounding (public, vendor SDK)

A representative agent SDK's `session.mcp_servers_loaded` event reports per-server provenance: `source` (`user` / `workspace` / `plugin` / `builtin`), `pluginName`, and `pluginVersion` — where `pluginVersion` is documented as "version of the plugin that supplied the effective MCP server config, only when source is plugin." A host projecting this models plugin-sourced servers as children of a plugin container, carrying `pluginName` as the container's `name`. `pluginVersion` describes the **plugin** (the container), not the individual server — but the container has no `version` field, so the value is dropped. A native `version` on `PluginCustomization` gives it a home.

(`transport` from that event is intentionally **not** in scope — AHP deliberately does not carry MCP transport; it's harness-internal.)

### Relationship to existing issues

- #303 (`argumentHint` on `SkillCustomization`) established the "native, type-specific field over a generic `_meta` bag" preference; this follows the same principle for `version`.
- Companion ask: #316 hoists a generic `_meta` slot onto `CustomizationBase` for provider-specific data with no first-class home. The two are complementary — `version` is deliberately a native field here, not a `_meta` key, because a plugin version is a first-class attribute.

### Backwards compatibility

Purely additive and optional — a new optional field on `PluginCustomization`. Existing producers and consumers are unaffected until they populate or read it.

colbylwilliams:
Closes #315.

## What

Adds an optional, display-only `version?: string` to `PluginCustomization` (inherited by `ClientPluginCustomization`).

```ts
export interface PluginCustomization extends ContainerCustomizationBase {
  type: CustomizationType.Plugin;
  /**
   * Version of the plugin, sourced from the
   * [Open Plugins](https://open-plugins.com/) manifest's optional
   * `version` field (semver, e.g. `"1.2.0"`). Absent when the manifest
   * declares no version — the field is optional there — or the source
   * has no version concept. Provenance / display only: the host neither
   * parses nor enforces it.
   */
  version?: string;
}
```

## Why

`PluginCustomization` mirrors an [Open Plugins](https://open-plugins.com/) package and already carries `name` via `CustomizationBase`, but not `version`. The Open Plugins spec lists `version` as an optional `plugin.json` metadata field — *"Semantic version (e.g., `"2.1.0"`). Used by tools to determine update availability"* — sitting right next to `name`. A host that resolved a plugin's version had nowhere to put it, so the value was dropped on the wire.

This follows the established **native, type-specific field over a generic `_meta` bag** precedent (`RuleCustomization.alwaysApply`/`globs`, `AgentCustomization.model`/`tools`, #303's `argumentHint`), because a plugin version is a first-class attribute of a plugin rather than incidental metadata. It is deliberately independent of #316, which hoists a generic `_meta` slot onto `CustomizationBase` for provider-specific data that lacks a first-class home — the two are complementary.

## Validation

I independently verified every assumption in the issue:

- **Open Plugins manifest** — confirmed against the [Open Plugins Specification v1.0.0](https://open-plugins.com/plugin-builders/specification): `version` is an **optional** metadata field (semver, "used by tools to determine update availability"), and `name` is the only required field. This maps cleanly onto AHP's required `name: string` + optional `version?`.
- **Mirror claim** — `PluginCustomization`'s doc comment already declares it "An Open Plugins plugin," and it carries `name` but no `version`.
- **Doctrine** — passes the design tests in `docs/guide/doctrine.md`: optional/additive (a minimal client can ignore it), durable state (lives in `SessionState.customizations`), host-authoritative, and not an agent implementation detail.
- **Reducers** — customizations are upserted as whole objects (`SessionCustomizationUpdated` / `SessionCustomizationsChanged`), so an optional field needs no reducer logic and no new fixtures; 100% branch coverage is unaffected.

## Ripple / scope

Matches the shape of a prior additive-optional-field change (#299, `contentSha256` on `ContentRef`):

- `types/channels-session/state.ts` — the field (hand-edited).
- Regenerated the five client mirrors (Go, Rust, Kotlin, Swift, TypeScript) and the JSON Schema via `npm run generate`.
- CHANGELOG entries added to the spec and every client, per the repo's changelog rules for a `types/**` change.
- `PROTOCOL_VERSION` intentionally **not** bumped.

## Testing

`npm run generate` and `npm run test` both run clean: typecheck, lint, `verify:release-metadata`, and `verify:changelog` all pass, and the full suite is **292/292**. (The `c8` coverage wrapper crashes locally on Node v26 via a yargs ESM/CJS incompat in its bootstrap — unrelated to this change; CI runs on a pinned Node.)

> [!NOTE]
> #316 is being implemented in parallel and also touches `types/channels-session/state.ts` and the CHANGELOGs. The changes are logically independent (different location in the file, complementary intent); a mechanical merge may be needed depending on land order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,10 @@ changes accumulate. Track in-flight protocol changes via PRs touching
   functionality.
 - `changeset/filesReviewedChanged` action for servers to update the `reviewed`
   flag of one or more changeset files.
+- Optional `version` field on `PluginCustomization` (inherited by
+  `ClientPluginCustomization`), carrying the plugin's semver sourced from the
+  Open Plugins manifest's optional `version`. Provenance / display only; absent
+  when the manifest declares no version or the source has no version concept.
 
 ## [0.5.2] — Unreleased
 
--- a/clients/go/CHANGELOG.md
+++ b/clients/go/CHANGELOG.md
@@ -35,6 +35,9 @@ Implements AHP 0.5.2.
   `Version`, optional `Title`), identifying the implementation and build behind
   either side of the handshake. Informational only — MUST NOT be used for
   feature detection.
+- Optional `Version` field on `PluginCustomization` (inherited by
+  `ClientPluginCustomization`), carrying the plugin's semver sourced from the
+  Open Plugins manifest. Provenance / display only.
 
 ### Changed
 
--- a/clients/go/ahptypes/state.generated.go
+++ b/clients/go/ahptypes/state.generated.go
@@ -2076,6 +2076,13 @@ type PluginCustomization struct {
 	// nothing.
 	Children []ChildCustomization `json:"children,omitempty"`
 	Type     CustomizationType    `json:"type"`
+	// Version of the plugin, sourced from the
+	// [Open Plugins](https://open-plugins.com/) manifest's optional
+	// `version` field (semver, e.g. `"1.2.0"`). Absent when the manifest
+	// declares no version — the field is optional there — or the source
+	// has no version concept. Provenance / display only: the host neither
+	// parses nor enforces it.
+	Version *string `json:"version,omitempty"`
 }
 
 // A {@link PluginCustomization} as published by a client. Extends the
@@ -2124,6 +2131,13 @@ type ClientPluginCustomization struct {
 	// nothing.
 	Children []ChildCustomization `json:"children,omitempty"`
 	Type     CustomizationType    `json:"type"`
+	// Version of the plugin, sourced from the
+	// [Open Plugins](https://open-plugins.com/) manifest's optional
+	// `version` field (semver, e.g. `"1.2.0"`). Absent when the manifest
+	// declares no version — the field is optional there — or the source
+	// has no version concept. Provenance / display only: the host neither
+	// parses nor enforces it.
+	Version *string `json:"version,omitempty"`
 	// Opaque version token used by the host to detect changes.
 	Nonce *string `json:"nonce,omitempty"`
 }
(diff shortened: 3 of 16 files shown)


connor4312 (approved):
I'm good if we want to add the open plugin manifest metadata here (description/author)

anthonykim1 (approved):

PR #332, opened 2026-07-10
chat: add turn timing metadata

justschen:
- Carries authoritative start and end timestamps through turn lifecycle actions so clients can render consistent timing information.
- Persists turn start times and derived durations in state while preserving compatibility with older completed turns.
- Keeps reducer behavior aligned across TypeScript, Rust, Kotlin, Swift, and Go, including invalid timestamp handling.
- Regenerates schemas and client wire types and extends shared conformance fixtures.

References microsoft/vscode#325061

(Commit message generated by Copilot)


types/channels-chat/actions.ts
```diff
@@ -343,6 +345,8 @@ export interface ChatTurnCompleteAction {
   type: ActionType.ChatTurnComplete;
   /** Turn identifier */
   turnId: string;
+  /** ISO 8601 timestamp when this turn completed. */
+  endedAt: string;
```
connor4312: Thinking on it more I recommend actually having a duration in the complete/cancelled/errors states rather than endedAt, because

1. It avoids date parsing in reducers which seems a bit sketch (and required extra dependencies for Rust)
2. Some of these actions are client dispatchable, some of them are server dispatchable. It's possible the clocks might not be in sync, and each issuing absolute timestamps could cause an endedAt to be before a startedAt. Using a duration avoids this. It's still possible that startedAt of subseqent turns could be non-monotonic because of different clocks on various clients, but this is not as big of a concern imo.

roblourens: Would it help if we could keep it off a client-dispatchable action, and the host would be in charge of updating timestamps as state?

Why would we need to parse the date in a reducer?

connor4312 (approved):
lgtm, fyi @colbylwilliams

PR #333, opened 2026-07-13
chat: add judge confirmation rationale

justschen:
## Summary

Adds model-judge confirmation rationale for the SDK auto-approval flow, mirroring microsoft/vscode#325547.

- Add typed judge rationale to `chat/toolCallReady`
- Persist the rationale in pending-confirmation state
- Propagate it through all client reducers
- Regenerate schemas and SDK types
- Add reducer fixtures and a changelog fragment


types/channels-chat/actions.ts
```diff
@@ -203,6 +204,8 @@ export interface ChatToolCallReadyAction extends ToolCallActionBase {
   toolInput?: string;
   /** Short title for the confirmation prompt (e.g. `"Run in terminal"`, `"Write file"`) */
   confirmationTitle?: StringOrMarkdown;
+  /** Why the tool requires user confirmation. */
+  confirmationReason?: ToolCallJudgeConfirmationReason;
```
connor4312: We already have `confirmed?: ToolCallConfirmationReason` down below, do we need a new property?

types/channels-chat/state.ts
```diff
@@ -934,6 +934,25 @@ export const enum ToolCallConfirmationReason {
...
+
+/**
+ * A model judge's explanation for requiring user confirmation.
+ *
+ * @category Tool Call Types
+ */
+export interface ToolCallJudgeConfirmationReason {
+  kind: ToolCallJudgeConfirmationReasonKind.Judge;
+  reason: string;
+}
```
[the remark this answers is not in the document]
justschen: Updated `reason` to `StringOrMarkdown` and regenerated the clients in d5bf842.

types/channels-chat/actions.ts
justschen: We do need a separate property here: `confirmed` represents a positive confirmation outcome, and its presence makes the reducer transition `chat/toolCallReady` directly to `running`. The judge rationale is emitted in the opposite case—when the judge requires user confirmation and the tool transitions to `pending-confirmation`. Reusing `confirmed` would conflate mutually exclusive states and invert its existing semantics.

justschen: ^ from copilot

i think we realistically could use the same thing, but for mental model maybe it's nice to keep it separate? or maybe rename it. it's more of a `denied` reason than a `confirmed` reason, since it's what we surfacce to the ux like the risk assessment

connor4312: Ah I see, sorry I misunderstood and made the wrong suggestion. 

1. Should we support 'judging' coming in async? If so, while we don't need an enum like `ToolCallJudgeConfirmationReasonKind` we may want to be able to have different states to indicate judgement is loading in.
2. Would we want some kind of `safety: number /* 0 = unsafe, 1 = safe */` hint so clients can the judgement with appropriate visuals?

justschen: 1. i think it's probably okay to not atm - i think it waits for the judgement to come back, then emits that the tool call is ready, so the confirmation comes in together. if we wanted to tho, it would be similar to our risk assessment (it shows a little spinner in the confirmation). i don't really know what the latency between tool actually being ready + judgement being ready is tho but i think CLI might have some insight in this
2. rn since it's just basically `1 = safe` and `2= unsafe`, i think we also don't need to since it's just binary of `yes` or `no`. similar to our own risk assessment, if we get different judgement levels (ie, `3= safe, but still delegate to user`), it can be cool for the ux to show different types of risk, but SDK seems to just be `yes` vs `no` rn. 

not 100% married to either decsions tho, so if you have preferences, we can do either or!

connor4312: I think we should support both. Even if the SDK doesn't expose it right now, async judgement seems like something we'd probably want in the future and multiple risk levels seems very in scope as well.

types/channels-chat/state.ts
```diff
@@ -1100,6 +1151,8 @@ export interface ToolCallPendingConfirmationState extends ToolCallBase, ToolCall
   status: ToolCallStatus.PendingConfirmation;
   /** Short title for the confirmation prompt (e.g. `"Run in terminal"`, `"Write file"`) */
   confirmationTitle?: StringOrMarkdown;
+  /** Why the tool requires user confirmation. */
+  confirmationReason?: ToolCallJudgeConfirmationReason;
```
connor4312: Let's rename from `confirmationReason` to something like `riskAssessment`, and likewise update the interfaces/enums. Then I think this should be good.

justschen: oh sorry i meant to do that yeah! sounds good thanks!

connor4312 (dismissed):

connor4312 (approved):

PR #337, opened 2026-07-16
spec: multiroot sessions (draft for feedback)

sandy081:
> **Draft for design feedback — not ready to merge.** Sharing the multiroot-sessions spec + proposal to gather review before further work.

## What this is

A first-class **multiroot sessions** proposal for AHP: a session can grant its agent tool access to **multiple working directories** (equal peers, no privileged "primary"); a **chat** operates on a **subset** of those; and **file changes group by directory**.

The plan/spec document is **[`docs/proposals/multiroot-sessions.md`](docs/proposals/multiroot-sessions.md)** — start there. It covers the problem, mental model, what-it-is/isn't, a concrete **protocol-surface summary** (change table + TypeScript signatures in §8), design decisions, and open questions.

## Protocol changes (all additive & optional, target `0.6.0`)

- **Capability** — `AgentCapabilities.multipleWorkspaceFolders` gates the whole feature.
- **Session** — `CreateSessionParams.workingDirectories` + `SessionMetadata.workingDirectories` (→ `SessionState`/`SessionSummary`); singular `workingDirectory` **deprecated**.
- **Commands** — `addWorkspaceFolder` / `removeWorkspaceFolder` (session) and `addChatWorkspaceFolder` / `removeChatWorkspaceFolder` (chat subset); removal is idempotent, modelled as *reconfigure-to-reduced-set*.
- **Chat** — `ChatState`/`ChatSummary`/`CreateChatParams.workingDirectories` (subset ⊆ session); singular **deprecated**.
- **Changes** — `Changeset.workingDirectory` + `'directory'` `changeKind`; one changeset per directory instead of arrays-of-arrays.

Old clients that never set the fields / send the commands behave exactly as today. See §8 of the proposal for the full change table and signatures.

## Open questions for reviewers

- Keep `workingDirectory` **deprecated** (as here) vs. hard-remove it (pre-1.0 allows it)?
- Make `resolveSessionConfig` / `sessionConfigCompletions` multiroot-aware? (currently keep their singular `workingDirectory` context.)
- Do session lists need per-directory change rollups (`ChangesSummary.byDirectory`)?

## Validation

`npm test` green — typecheck + lint + verify-release-metadata + verify-changelog + 316 reducer/message tests, 100% branch coverage. Schemas + all five clients regenerated.

## Note

Branch is ~10 commits behind `origin/main`; will rebase before any real merge. Draft-only for feedback.


types/channels-chat/commands.ts
```diff
@@ -52,3 +64,67 @@ export interface CreateChatParams extends BaseParams {
...
+ * directory already in the chat's set is a no-op that still returns the current
+ * full set.
+ *
+ * @category Commands
+ * @method addChatWorkspaceFolder
+ * @direction Client → Server
+ * @messageType Request
+ * @version 1
+ */
+export interface AddChatWorkspaceFolderParams extends BaseParams {
```
connor4312: These should probably be actions state actions instead of commands, have your agent follow the "Actions that mutate a keyed collection" instruction in general-instructions

types/channels-chat/state.ts
```diff
@@ -60,13 +60,26 @@ export interface ChatState {
...
-   * {@link SessionState.workingDirectory | the session's working directory}.
-   * Hosts MAY override this for individual chats — for example, to give a
-   * subordinate chat its own git worktree so multiple chats in a session can
-   * make independent edits that the orchestrator later merges back.
+   * @deprecated Use {@link workingDirectories} instead. Retained as a
+   * single-directory shorthand for backwards compatibility. When
+   * `workingDirectories` is present this field reflects its first entry so
+   * older clients continue to see a usable value.
    */
   workingDirectory?: URI;
```
connor4312: This week we're doing a breaking change in 0.6.0 -- we can do a hard swap unless this gets deferred to next week

types/channels-root/state.ts
```diff
@@ -126,6 +134,15 @@ export interface MultipleChatsCapability {
...
 }
 
+/**
+ * Options for the {@link AgentCapabilities.multipleWorkspaceFolders} capability.
+ * Currently carries no sub-options; presence of an empty object `{}` is the
+ * entire signal.
+ *
+ * @category Root State
+ */
+export interface MultipleWorkspaceFoldersCapability {}
```
connor4312: Maybe have a `immutablePrimary?: boolean` to indicate whether the first directory can be modified, as discussed

types/channels-session/commands.ts
```diff
@@ -117,6 +137,75 @@ export interface CreateSessionParams extends BaseParams {
...
+ * "primary". Adding a directory that is already in the set is a no-op that
+ * still returns the current full set.
+ *
+ * @category Commands
+ * @method addWorkspaceFolder
+ * @direction Client → Server
+ * @messageType Request
+ * @version 1
+ */
+export interface AddWorkspaceFolderParams extends BaseParams {
```
connor4312: Same action swap here

CHANGELOG.md
```diff
@@ -27,6 +27,59 @@ changes accumulate. Track in-flight protocol changes via PRs touching
 
 Spec version: `0.6.0`
 
+### Added
```
connor4312, line 34: generate a fragment instead please, see AGENTS.md

types/channels-changeset/state.ts
```diff
@@ -63,6 +66,18 @@ export interface Changeset {
...
+   * within a single directory of a multiroot session. MUST be one of the
+   * session's {@link SessionState.workingDirectories}.
+   *
+   * A host with multiple working directories groups changes by directory by
+   * advertising one changeset per directory, each carrying its
+   * `workingDirectory` — rather than nesting changes as arrays-of-arrays.
+   * Omit for changesets that are not directory-scoped (e.g. a single-directory
+   * session, or a session-wide roll-up spanning every directory).
+   */
+  workingDirectory?: URI;
```
roblourens: Would it work to say that Changesets always just cover all directories? Then you can have one changeset which covers all changes made, and it should be easy to see which dir a particular change is part of, based on which dir the file is in.

Also I'm not sure how the "Other files" work today, we can already track changes outside of a working dir

types/channels-chat/state.ts
roblourens: I think we just need to make sure we have a working implementation somewhere before we merge the protocol changes

types/channels-changeset/state.ts
connor4312: Yea, that also avoids future questions about nested git repos etc., I would drop this unless we have an experience we struggle to drive without it

sandy081: I'd actually push to keep this and go further — make server-side grouping a normative **MUST** (while the field stays optional). My reasoning is that grouping changes by directory is a **host** responsibility, not a client one.

Per AHP's [doctrine](docs/guide/doctrine.md), AHP is a *"client-facing presentation model"* giving clients *"display-ready session state ... without requiring the client to understand a particular filesystem model,"* with the *"host [owning] the authoritative state."* Deriving a file's directory by prefix-matching its URI against the session's working directories **is** a filesystem model — and if the server doesn't do it, every client (VS Code, CLIs, web, mobile) re-implements the same grouping independently and inconsistently. That's exactly the semantic work AHP is meant to centralize in the host.

And the client can't even do it correctly: only the host knows the real VCS boundaries. A file under `repo-a/` may actually live in a nested `repo-a/vendor/repo-c/` git repo; longest-prefix matching gets that wrong, and the client can't see symlinks, submodules, or worktrees. The host is the only party that knows which root a change truly belongs to — so grouping belongs where that knowledge is.

So rather than dropping it, I've changed the spec to say: **a host with multiple working directories MUST emit one changeset per directory** (each with `workingDirectory` set). The field stays **optional** only so the cases you both raised still work — a single-directory session, and out-of-tree / aggregate changesets that intentionally span or sit outside the working dirs (Rob's "Other files"). So nothing we can track today is lost, a minimal client can still ignore the field, and it's gated by `multipleWorkspaceFolders`.

Bonus: it also fits the existing catalogue (the host already advertises `session`/`turn`/`compare-turns`/`branch`/`uncommitted` views — `directory` is one more server-owned slice), and gives per-repo operations like `create-pr`/`commit` a natural scope, since those are inherently per-repo with sibling repositories.

On nested repos specifically — I'd argue an explicit `workingDirectory` on the changeset *resolves* the ambiguity rather than creating it: the host states which root owns the changeset instead of leaving clients to guess.

types/channels-session/commands.ts
sandy081: Done — converted to state actions. `addWorkspaceFolder`/`removeWorkspaceFolder` are now the client-dispatchable `session/workingDirectorySet` / `session/workingDirectoryRemoved` actions (keyed by the directory URI, idempotent), with reducers, `ACTION_INTRODUCED_IN` entries, and full-branch conformance fixtures.

types/channels-chat/commands.ts
sandy081: Done — same swap at the chat level: `chat/workingDirectorySet` / `chat/workingDirectoryRemoved` actions replace the commands.

types/channels-chat/state.ts
sandy081: Done — hard-removed the singular `workingDirectory` from `CreateSessionParams`, `SessionMetadata`, `ChatState`, and `ChatSummary` (no deprecated shorthand), and updated the hand-written client reducers/mergers in every language. Agreed on the working-implementation gate — that's happening on the vscode side before this merges.

types/channels-root/state.ts
sandy081: Added `immutablePrimary?: boolean` to `MultipleWorkspaceFoldersCapability`: when set, index 0 is a fixed process root that clients MUST NOT remove/reorder, while the rest stay equal peers. The `workingDirectoryRemoved` action reducer/doc notes a host MAY decline to apply a removal in that case.

CHANGELOG.md
sandy081: Done — moved the entries out of the changelogs into `docs/.changes/20260716-multiroot-sessions.json`.

types/channels-root/state.ts
```diff
@@ -110,6 +110,17 @@ export interface AgentCapabilities {
...
+   * The session's agent can be granted tool access to more than one working
+   * directory. The directories are treated as equal peers except where the
+   * agent advertises {@link MultipleWorkspaceFoldersCapability.immutablePrimary}
+   * (some backends pin their first directory as a fixed process root).
+   *
+   * When absent, clients MUST NOT mutate a session's or chat's working-directory
+   * set and MUST NOT set more than one entry in
+   * {@link CreateSessionParams.workingDirectories}.
+   */
+  multipleWorkspaceFolders?: MultipleWorkspaceFoldersCapability;
```
DonJayamanne: Should we rename this to `multipleWorkingDirectories` instead?
Everywhere else we've used the word `workingDirectory`, but here we're using `workspaceFolder` (which feels more of the vscode terminology)

types/channels-changeset/state.ts
roblourens: I don't understand the point about VCS boundaries- we care about matching the file to a directory, which is a list we already have, not a git repo.

Also, I'm not actually sure that the client would typically care to group by working directory anyway, does it matter? I think I'd render a tree for the session and call it a day.

But if you want to group them, understanding URIs at this level doesn't seem like it's making wild assumptions about the remote.

types/channels-root/state.ts
sandy081: Good catch — renamed `multipleWorkspaceFolders` → `multipleWorkingDirectories` (and `MultipleWorkspaceFoldersCapability` → `MultipleWorkingDirectoriesCapability`) so the whole feature stays on the `workingDirectory` vocabulary. The VS Code-flavoured "workspace folder" term is gone.

types/channels-changeset/state.ts
sandy081: You've convinced me — dropped `Changeset.workingDirectory` (and the `'directory'` changeKind). You're right that matching a file to one of the session's directories is a simple client-side operation against a list we already have, not a VCS-boundary problem, so the "clients can't do it" argument doesn't hold.

There's also a modelling flaw I'd missed: a changeset can *span* directories (a per-turn diff touches several), so a single-directory scalar can't represent the common case anyway. So the new shape is: changesets stay cross-cutting; a client that wants a per-directory view groups the files itself against `workingDirectories`; and a host MAY additionally advertise dedicated per-directory changesets as extra catalogue entries (the `changesets` list is already unbounded — no schema change needed). If we later want machine-readable per-directory entries, a `{workingDirectory}` uriTemplate variable is the natural follow-up.

sandy081: cc @lszomoru

lszomoru: @sandy081, I would like to point out what while some changeset may span multiple working directories (ex: last turn changes), some changesets are very much scoped to a git repository (ex: branch changes). Branch changes which currently is the default changeset for a session, compares `branchA` with `main` with the branch information even present in the changeset description which is surfaced in the UI. In light of that I would like to suggest to reconsider the optional `workingDirectory` on the changeset.

connor4312 (dismissed):

connor4312 (dismissed):

connor4312:
0.7.0 is fine here, this is a breaking change

sandy081:
## Update: `requiresPrimary` capability + explicit `primaryWorkingDirectory` (session & chat)

Pushed in `148c716b`. This reworks the "primary directory" concept based on offline discussion with @TylerLeonhardt 

**What changed**

1. **Renamed the capability flag `immutablePrimary` → `requiresPrimary`.** It now means *"the agent needs one of its working directories designated as the primary"* (a distinguished root it centers on — process root, default for relative paths), rather than *"index 0 is a fixed process root"*.

2. **Primacy is now explicit, never positional.** Added a `primaryWorkingDirectory: URI` field so the primary is designated explicitly rather than inferred from array index. This keeps the "directories are equal peers, order carries no meaning" invariant intact — removing/reordering entries can't silently repromote a new primary.
   - `CreateSessionParams.primaryWorkingDirectory` (⊆ `workingDirectories`) → reported on `SessionState.primaryWorkingDirectory` / `SessionSummary`.
   - `CreateChatParams.primaryWorkingDirectory` (⊆ the chat's dirs) → reported on `ChatState`/`ChatSummary.primaryWorkingDirectory`. A chat **inherits the session's primary** when it omits its own.

3. When `requiresPrimary` is set, a client SHOULD supply `primaryWorkingDirectory`; a host MAY reject creation that omits it

types/channels-session/reducer.ts
```diff
@@ -226,6 +226,30 @@ export function sessionReducer(state: SessionState, action: SessionAction, log?:
...
+      const list = state.workingDirectories;
+      if (!list) {
+        return state;
+      }
+      const idx = list.indexOf(action.directory);
+      if (idx < 0) {
+        return state;
+      }
+      const updated = list.slice();
+      updated.splice(idx, 1);
```
dmitrivMS, line 249: AI Review: Could this removal be rejected unless related state is updated transactionally? As written it can leave `primaryWorkingDirectory` pointing outside the session set and existing chat subsets referencing a directory the session no longer owns. There is also no action for reassigning the primary before removing it.

types/version/registry.ts
```diff
@@ -34,7 +34,7 @@ export const PROTOCOL_VERSION = '0.6.1';
  * `scripts/verify-release-metadata.ts`.
  */
 export const SUPPORTED_PROTOCOL_VERSIONS: readonly string[] = Object.freeze([
-  '0.6.1',
+  '0.7.0',
```
dmitrivMS, line 37: AI Review: Can clients built from this source still truthfully advertise `0.6.0` fallback? `CreateSessionParams` has removed the singular `workingDirectory`, so after negotiating 0.6 this client no longer has a way to construct that version's session-creation wire shape.

types/channels-chat/reducer.ts
```diff
@@ -382,6 +382,30 @@ export function chatReducer(state: ChatState, action: ChatAction, log?: (msg: st
     case ActionType.ChatActivityChanged:
       return { ...state, activity: action.activity };
 
+    // ── Working Directories ───────────────────────────────────────────────
+
+    case ActionType.ChatWorkingDirectorySet: {
+      const list = state.workingDirectories ?? [];
```
dmitrivMS, line 388: AI Review: When `workingDirectories` is absent, the state contract says this chat inherits the full session set. Here absence is treated as an empty set, so adding one directory narrows the inherited set to a singleton, while the removal case below cannot remove an inherited directory. How can these actions preserve the documented inheritance semantics without access to the session state?

clients/typescript/src/client/hosts/runtime.ts
```diff
@@ -879,7 +879,7 @@ function applySummaryChange(
   if (changes.activity !== undefined) merged.activity = changes.activity;
   if (changes.modifiedAt !== undefined) merged.modifiedAt = changes.modifiedAt;
   if (changes.project !== undefined) merged.project = changes.project;
-  if (changes.workingDirectory !== undefined) merged.workingDirectory = changes.workingDirectory;
+  if (changes.workingDirectories !== undefined) merged.workingDirectories = changes.workingDirectories;
```
dmitrivMS, line 882: AI Review: Should this also merge `changes.primaryWorkingDirectory`? Without it, the cached session summary can retain a stale primary. The Rust and Swift host runtime summary mergers have the same omission.

clients/go/ahptypes/common.go
```diff
@@ -83,8 +83,10 @@ type PartialChatSummary struct {
 	ModifiedAt *string `json:"modifiedAt,omitempty"`
 	// How this chat came into existence
 	Origin *ChatOrigin `json:"origin,omitempty"`
-	// Optional per-chat working directory.
-	WorkingDirectory *URI `json:"workingDirectory,omitempty"`
+	// The subset of the session's working directories this chat uses.
+	WorkingDirectories []URI `json:"workingDirectories,omitempty"`
+	// The chat's primary working directory.
```
dmitrivMS, line 88: AI Review: `omitempty` collapses a present-but-empty slice into an absent field, but the chat contract distinguishes explicit empty (`no working-directory access`) from absent (`inherit the session set`). Can the Go representation preserve that distinction, or should the protocol drop the explicit-empty semantic?

types/channels-session/reducer.ts
```diff
@@ -226,6 +226,30 @@ export function sessionReducer(state: SessionState, action: SessionAction, log?:
       return { ...state, activeClients: updated };
     }
 
+    // ── Working Directories ─────────────────────────────────────────────
+
+    case ActionType.SessionWorkingDirectorySet: {
+      const list = state.workingDirectories ?? [];
+      if (list.includes(action.directory)) {
```
dmitrivMS, line 233: AI Review: What defines URI identity for this set? Every reducer currently uses raw string equality, so equivalent URI spellings can be duplicated or fail removal across clients. The protocol should define wire-level equality or provide conformance fixtures for the intended behavior.

DonJayamanne (dismissed):

connor4312 (approved):

types/channels-changeset/state.ts
sandy081: @lszomoru you have the best understanding of how change sets should be represented. I do not have objections to include this. Lets also get @roblourens on same page.

Thanks

Issue #323, opened 2026-07-08
Consider a lightweight progress content block for streaming non-terminal-backed shell output


anthonykim1:
Follow-up from #314, discussion thread: https://github.com/microsoft/agent-host-protocol/pull/314#discussion_r3540133509

Rob asked whether shell exit content should be separated so terminal output can stream. We resolved this for the case where a full `ahp-terminal:` channel backs the command: a host emits a live `ToolResultTerminalContent` block while it runs, then a `ToolResultTerminalCompleteContent` block referencing the same `resource` when it exits.

That doesn't cover the lighter case — a plain shell child process with no subscribable terminal channel. There's currently no way for a host to stream partial output/progress for that kind of execution before the final `TerminalComplete` summary lands.

**Problem**

- `ToolResultTerminalContent` implies a full terminal channel (cols/rows, command detection, etc.) — heavier than some shell tools want just to report incremental output.
- Non-terminal-backed shells only get one shot: the final `preview` in `TerminalComplete`, after the command has already exited.

**Proposal (not fully speced yet)**

- A lighter content type for incremental output from a running command that isn't backed by a full terminal channel. Naming isn't settled — `ToolResultShellProgressContent` vs `ToolResultTerminalProgressContent` vs something else — depends on whether we want this to read as shell-specific or as part of the same terminal-content family as `ToolResultTerminalContent`/`ToolResultTerminalCompleteContent`.
- Needs to stay clearly distinct from `ToolResultTerminalCompleteContent`, which is specifically the post-exit record.

**Sequencing**

We should land #322 (PTY / non-PTY capability detection) before this. Once clients can tell whether an execution is PTY-backed vs a plain child process, that same signal likely informs whether/how a progress block should behave — don't want to design streaming semantics twice.

**Related**

- #314 — where this was raised and partially resolved (terminal-backed case)
- #322 — capability detection for PTY vs non-PTY terminal execution (should land first)

PR #352, opened 2026-07-21
Add isPty and fold TerminalComplete into ToolResultTerminalContent

### LINKED ISSUE #322: Track terminal execution capabilities for PTY and non-PTY command output
AHP currently exposes terminal output through `ToolResultTerminalContent` / `ahp-terminal:` and command completion metadata through `ToolResultTerminalCompleteContent`.

During discussion in #314, we intentionally avoided specifying whether an `ahp-terminal:` channel is always PTY-backed or may also represent non-interactive command execution, because clients currently do not have a protocol-supported way to discover the backing behavior.

We should discuss whether AHP needs terminal or command-execution capability metadata so clients can make rendering and interaction decisions without guessing from implementation details.

## Questions to discuss

- Should `ahp-terminal:` always mean PTY-backed terminal, or can it represent a broader command execution surface?
- If non-PTY command output can use terminal channels, what capabilities should clients be able to inspect?
- Do we need fields such as `supportsInput`, `supportsResize`, `supportsVT`, stdout/stderr separation, or other execution/rendering hints?
- Should this capability metadata live on `TerminalState`, a command execution state shape, or a future channel kind?
- How should clients render retained command completion metadata when no live terminal channel exists?

## Context

- #314 added `ToolResultTerminalCompleteContent` as retained command completion metadata.
- It optionally references an `ahp-terminal:` resource when live output was exposed.
- We removed wording that implied terminal channels may not be PTY-backed until there is a concrete capability-discovery story.

## Goal

Define a protocol-level way for clients to determine what terminal or command execution behaviors are available, rather than assuming PTY semantics or encoding host implementation details in naming.

anthonykim1:
Resolves: https://github.com/microsoft/agent-host-protocol/issues/322, https://github.com/microsoft/agent-host-protocol/issues/323
Part of: https://github.com/microsoft/vscode/issues/324825

- Add optional `isPty` to `ToolResultTerminalContent` so a client knows whether a terminal is PTY-backed before subscribing or initializing an emulator.
- Add matching `isPty` to `TerminalState` so the channel is self-describing regardless of how a client discovered it; `isPty: false` means plain-text output and clients do not need to parse VT sequences.
- Fill `exitCode`, `preview`, and `truncated` in on the terminal block in the completed result, so clients that did not subscribe to the terminal still get the command outcome. This records the command's exit, not the terminal's — the terminal may keep running afterwards.
- Remove `ToolResultTerminalCompleteContent` (breaking, riding the 0.7.0 window): its completion data now lives directly on the terminal block it previously correlated with, so clients no longer pair two content blocks per command.
- Leave `cwd` off the completion data deliberately — it can be added back additively if a client needs to display it at the top-level terminal part.
- Keep `resource` and `title` required. Per review discussion, non-pty shell runs will be exposed as resource-backed terminals (`isPty: false`) whose output streams over the terminal channel, so clients receive output only when they subscribe.
- Add round-trip fixtures `030-terminal-content-ispty` and `031-terminal-content-completed`, stage `added`/`removed` change fragments, and regenerate the JSON schemas and Go/Rust/Kotlin/Swift/TypeScript clients.

-----------------------------------------
Inspirations from:

- The completion fields are relocated semantically from the removed [`ToolResultTerminalCompleteContent`](https://github.com/microsoft/agent-host-protocol/blob/0960c7e1b74860bb055617839aa49ff976b5c39d/types/channels-chat/state.ts#L1416-L1435) — same `exitCode`/`preview`/`truncated` shapes and doc intent, moved onto the terminal block.
- The `isPty` capability-flag placement on `TerminalState` follows the existing [`supportsCommandDetection`](https://github.com/microsoft/agent-host-protocol/blob/0960c7e1b74860bb055617839aa49ff976b5c39d/types/channels-terminal/state.ts#L101-L109) pattern.
- The streaming path non-pty terminals will use already exists: [`terminal/data` appends into `TerminalState.content`](https://github.com/microsoft/agent-host-protocol/blob/0960c7e1b74860bb055617839aa49ff976b5c39d/types/channels-terminal/actions.ts#L13-L32), so late subscribers reconstruct output from channel state.
- The gap this feeds into is VS Code's tool-name sniffing to decide whether partial results are terminal output: [`isShellTool`](https://github.com/microsoft/vscode/blob/d5862406ee909d95ad04d8dac35c3a8646b886a6/src/vs/platform/agentHost/node/copilot/copilotToolDisplay.ts#L469-L474) — the protocol now labels terminal resources instead.


types/channels-chat/state.ts
```diff
@@ -1384,19 +1384,35 @@ export interface ToolResultFileEditContent extends FileEdit {
...
+  resource?: URI;
   /** Display title for the terminal content */
-  title: string;
+  title?: string;
+  /**
+   * Inline snapshot of output produced so far. A replacement snapshot, not a
+   * delta: each `chat/toolCallContentChanged` action supersedes the previous
+   * snapshot.
+   */
+  output?: string;
```
connor4312: If I'm just emitting an output snapshot and not a stream, wouldn't I use `ToolResultTerminalCompleteContent`?

types/channels-chat/state.ts
```diff
@@ -1384,19 +1384,35 @@ export interface ToolResultFileEditContent extends FileEdit {
...
+  /** Whether this terminal-style resource is backed by a pseudoterminal. */
```
connor4312: Should note here and on the state property that isPty:false = clients should not need to parse VT codes

types/channels-chat/state.ts
anthonykim1: `ToolResultTerminalCompleteContent` should be used only when we know for sure command has exited/completed and there would be no further output from particular command.
* I do see this creating confusion with `Preview of the command's output, if available` from `ToolResultTerminalCompleteContent` though. 🧐
* The name `preview` for `ToolResultTerminalCompleteContent` is meant to explain that there would be no further output, but the output may have been "truncated" bc its too long, etc.


Starting by being more clear on wording: [dd79bd5](https://github.com/microsoft/agent-host-protocol/pull/352/commits/dd79bd5dedd8d0c08526524577df5cbbf631ef79) ,curious what you think


Should we just have `delta` approach in `ToolResultTerminalContent` rather than a snapshot? But considering `ToolResultTerminalContent` as a "state" perhaps snapshot makes more sense.

types/channels-chat/state.ts
anthonykim1: Resolved via: [mention non pty do not have to parse VT sequences](https://github.com/microsoft/agent-host-protocol/pull/352/commits/c4f3f7bb20c71b1b40b7d3649bb65b6bf02b5805)

types/channels-chat/state.ts
```diff
@@ -1384,19 +1384,40 @@ export interface ToolResultFileEditContent extends FileEdit {
...
+ * {@link ToolResultTerminalCompleteContent} block, which retains completion
+ * metadata such as the exit code.
  *
  * @category Tool Result Content
  */
 export interface ToolResultTerminalContent {
   type: ToolResultContentType.Terminal;
   /** Terminal URI (subscribable for full terminal state) */
-  resource: URI;
+  resource?: URI;
```
anthonykim1: Concern I have is that making `resource` and `title` optional in `ToolResultTerminalContent` would be a "breaking" change.

types/channels-chat/state.ts
```diff
@@ -1397,34 +1401,18 @@ export interface ToolResultTerminalContent {
...
-  /** Working directory where the command was executed */
-  cwd?: URI;
-  /** Preview of the command's output, if available */
+  /**
+   * Preview of the command's output, for clients that are not subscribed
+   * to the terminal or that arrive after it is disposed
+   */
   preview?: string;
   /** Whether `preview` is known to be incomplete or truncated */
   truncated?: boolean;
```
connor4312, line 1431: I suggest grouping these under a `result?:` object. The presence of that object can signal that the command is complete, and then exit/preview/truncated can be filled in. We should also note that isPty affects the interpretation of `preview`

anthonykim1: addressed them here: [4c2e101](https://github.com/microsoft/agent-host-protocol/pull/352/commits/4c2e10105f13b1de659d3f253f8ad05fc48036a8)

connor4312 (approved):

PR #374, opened 2026-07-31
chat: lazy-load final tool input

roblourens:
## Summary

- make partial arguments and invocation-message updates independently optional on `chat/toolCallDelta`, so hosts may send either or both
- let hosts choose inline strings or `ContentRef` values for final `toolInput`
- keep edited input action-only: reducers replace inline input, while referenced input is updated by the host before echoing confirmation
- regenerate schemas and all five clients, with cross-language reducer fixtures

## Validation

- `npm run test` (354 tests, 100% reducer coverage)
- TypeScript client typecheck and tests (65 tests)
- Rust formatting, Clippy, workspace tests, and doc tests
- Swift package tests (115 tests)
- Kotlin Gradle tests
- Go tests
- second `npm run generate` pass and `git diff --check`

(Written by Copilot)

### DIFF
--- a/types/channels-chat/actions.ts
+++ b/types/channels-chat/actions.ts
@@ -5,7 +5,7 @@
  */
 
 import { ActionType } from '../common/actions.js';
-import type { StringOrMarkdown, ErrorInfo, FileEdit, UsageInfo, URI } from '../common/state.js';
+import type { ContentRef, StringOrMarkdown, ErrorInfo, FileEdit, UsageInfo, URI } from '../common/state.js';
 import type { McpAuthRequirement } from '../channels-session/state.js';
 import type {
   Message,
@@ -173,10 +173,8 @@ export interface ChatToolCallStartAction extends ToolCallActionBase {
  */
 export interface ChatToolCallDeltaAction extends ToolCallActionBase {
   type: ActionType.ChatToolCallDelta;
-  /** Partial parameter content to append */
-  content: string;
-  /** Updated progress message */
-  invocationMessage?: StringOrMarkdown;
+  /** Updated display-ready progress message */
+  invocationMessage: StringOrMarkdown;
 }
 
 /**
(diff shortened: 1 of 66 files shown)


types/channels-chat/actions.ts
```diff
@@ -211,8 +209,8 @@ export interface ChatToolCallReadyAction extends ToolCallActionBase {
   intention?: string;
   /** Message describing what the tool will do or what confirmation is needed */
   invocationMessage: StringOrMarkdown;
-  /** Raw tool input */
-  toolInput?: string;
+  /** Reference to the final raw tool input, readable with `resourceRead` */
+  toolInput?: ContentRef;
```
connor4312: Thoughts on allowing this to be a `string | ContentRef` and leaving it to the host to decide?

roblourens: How would a host decide one way or another?

roblourens: It seems like it would be the client that cares- based on whether it wants to render raw tool args, or whether it is very bandwidth-sensitive

connor4312:
@roblourens streaming tool input is used for MCP apps to progressive rendering (e.g. in Claude https://claude.com/blog/claude-builds-visuals at 0:30 in their video; FastMCP supports this too) so I'm not sure we should just remove the ability.

roblourens:
Ah, so for some types of tools it's unavoidable and we want the host to decide that? Makes sense

types/channels-chat/actions.ts
roblourens: Updated this to `string | ContentRef`, with the host choosing the representation. I also restored optional partial argument chunks on `chat/toolCallDelta`; `invocationMessage` remains required. (Written by Copilot)

dmitrivMS (approved):

connor4312 (approved):

PR #375, opened 2026-07-31
types: rename ResourceReponsePart to ResourceResponsePart

roblourens:
## Rename `ResourceReponsePart` → `ResourceResponsePart`

The interface name misspelled "Response". The Rust and Go generators already carried per-language rename entries to correct it on output:

```ts
// scripts/generate-rust.ts (before)
{ name: 'ResourceReponsePart', omitDiscriminants: true, rustName: 'ResourceResponsePart' },
```

```ts
// scripts/generate-go.ts (before)
{ name: 'ResourceReponsePart', goName: 'ResourceResponsePart' },
```

…while TypeScript, Kotlin and Swift emitted the typo verbatim. So a single protocol type had **two different names across the five clients**, and consumers of the TS/Kotlin/Swift clients had to spell it wrong to name the type.

This corrects the canonical spelling in `types/channels-chat/state.ts` and deletes both now-redundant generator renames, so the name is derived consistently everywhere instead of being patched per-language. It also updates two hand-written references the generators don't reach — the `AHPApp` Swift sample app (which imports `AgentHostProtocol`, so it would otherwise no longer compile) and the state-model guide. Neither is covered by CI.

**No wire change** — the discriminant value stays `contentRef` and the type is structurally identical. Source-breaking only for TypeScript, Kotlin and Swift consumers that reference the type by name; Rust and Go consumers already saw the corrected spelling.

Worth doing before the 1.0 freeze, since renaming a published type afterwards is a major-version change.

## Verification

- `npm run generate` — no drift; zero remaining occurrences of the typo across `types/`, all five clients, the JSON Schemas and the docs
- `npm test` — 378 pass, reducer branch coverage at 100%
- `cargo test --workspace` and `go build ./... && go test ./...` — both green

<sub>(Written by Copilot)</sub>

### DIFF
--- a/types/channels-chat/state.ts
+++ b/types/channels-chat/state.ts
@@ -1185,9 +1185,23 @@ interface ToolCallBase {
 interface ToolCallParameterFields {
   /** Message describing what the tool will do */
   invocationMessage: StringOrMarkdown;
-  /** Raw tool input */
-  toolInput?: string;
+  /**
+   * Final tool input.
+   *
+   * Referenced input is mutable until the tool call leaves
+   * `pending-confirmation`. When the client confirms with `editedToolInput`,
+   * the host MUST replace the resource contents before echoing the accepted
+   * confirmation action. Clients MUST NOT cache tool input across confirmation.
+   */
+  toolInput?: ToolInput;
 }
+
+/**
+ * Tool input represented inline or by reference.
+ *
+ * @category Tool Call Types
+ */
+export type ToolInput = string | ContentRef;
 
 /**
  * Tool execution result details, available after execution completes.
@@ -1222,7 +1236,7 @@ export interface ToolCallResult {
  */
 export interface ToolCallStreamingState extends ToolCallBase {
   status: ToolCallStatus.Streaming;
-  /** Partial parameters accumulated so far */
+  /** Partial parameters accumulated from tool-call deltas. */
   partialInput?: string;
   /** Progress message shown while parameters are streaming */
   invocationMessage?: StringOrMarkdown;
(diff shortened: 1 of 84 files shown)


connor4312 (dismissed):

types/channels-chat/state.ts
```diff
@@ -39,8 +39,16 @@ export interface ChatState {
...
+  /**
+   * Chat title.
+   *
+   * Absent means the chat has no title of its own and consumers SHOULD fall
+   * back to the owning {@link SessionState.title | session's title}. This is
+   * the normal case for a session's default chat, which typically has no
+   * identity separate from the session itself. Producers MUST NOT use an empty
+   * string to mean "inherit" — omit the field instead.
+   */
+  title?: string;
```
connor4312: Do we actually want this to be optional? A client will want to display chat tabs or navigation, and a title is pretty much needed for that, even if it's just a placeholder like "Subagent Session" issued by the agent host

roblourens: As I understand it, the special case here is the default chat. It wants to use the session title, today we set `title: ''` and fall back to the session title by convention. We can do that, or duplicate the title onto the default chat, or make title optional but implicitly required for other chats, or remove the title from the session, idk.

roblourens: I'd left this as a draft because I wasn't sure what I wanted

roblourens: We can definitely just leave it

connor4312: I'd probably leave it for now 🤷

connor4312 (approved):

PR #390, opened 2026-08-11
Add durable error response parts and turn resume

roblourens:
## Summary

- Represent turn failures as durable `ErrorResponsePart` entries in the response stream while retaining `TurnState.Error` as the top-level terminal signal.
- Let hosts mark the final error part with `resumable: true`; clients decide whether and how to present a resume affordance.
- Add the client-dispatchable `chat/turnResume` action, which reopens the latest resumable errored turn with the same turn ID, message, response parts, and usage—without adding a synthetic message or turn.
- Preserve prior errors when a resumed turn completes or fails again.

## Semantics

The host accepts `chat/turnResume` only when no turn is active, the requested turn is the latest turn, that turn is errored, and its final response part is resumable. Invalid or stale resume actions are rejected before provider side effects run.

Minimal clients can ignore `resumable` and continue rendering a coherent terminal error. Clients that support resume can render any appropriate affordance and dispatch `chat/turnResume` when the user invokes it.

## Validation

- Generated documentation, schemas, action-origin metadata, and Swift, Rust, Kotlin, TypeScript, and Go protocol types.
- `npm test`: 396 passing with 100% reducer branch coverage.
- TypeScript client tests pass.
- Go client tests pass.
- Rust client passes `cargo check --tests`.

(Written by Copilot)


connor4312 (dismissed):
this is looking nice now in general

connor4312 (dismissed):

benibenj (approved):

PR #393, opened 2026-08-12
automations: feat: add shared automations protocol

ulugbekna:
## Summary

Add shared, host-owned Automations to Agent Host Protocol.

The protocol represents:

- durable automation definitions and optimistic revisions;
- manual, schedule, and host-defined event triggers;
- task-level runs with lifecycle, blockers, artifacts, and cancellation;
- one or more linked ordinary AHP sessions;
- durable automation/run provenance on created sessions;
- host capabilities for execution lifetime, creation, schedules, cancellation,
  preview, and retained history;
- idempotent, cutover-safe imports from legacy schedulers.

Multiple clients connected to one authority can share one catalogue, scheduler,
run claim, and history instead of independently scheduling duplicate sessions.

### `ahp-automation:`

Owns the reusable definition:

- title and initial user message;
- provider, model, custom agent, working directories, and session config;
- enabled state and trigger collection;
- revision and next occurrence;
- retained run summaries and permitted operations.

### `ahp-automation-run:`

Owns one invocation:

- manual or trigger provenance;
- pending, running, blocked, completed, failed, or cancelled lifecycle;
- blocker summary;
- linked sessions and primary session;
- run-scoped artifacts and permitted operations.

Ordinary session/chat channels remain authoritative for transcripts, tools,
confirmations, and changes.

## Triggers and execution

- Schedule triggers use one portable five-field cron expression plus an IANA
  time zone.
- The grammar covers numeric/named fields, wildcards, ranges, lists, steps, and
  Unix day-of-month/day-of-week matching.
- Event triggers use authority-defined types, actions, and schema-defined
  configuration discovered through `listAutomationTriggerDefinitions`.
- `execution.lifetime` distinguishes interactive `hostLifetime` scheduling from
  independently managed authorities.
- `runAutomation.requestId` is a durable idempotency key.

## Cutover-safe imports

`createAutomation.import` carries a stable source, batch, and item identity.
It may also carry each schedule trigger's next unevaluated source occurrence.

- Imported definitions must be created disabled.
- The host persists supplied occurrences while the definition is disabled.
- After the client removes the legacy scheduler-visible copy and enables the
  host definition, overdue work follows the trigger's misfire policy.
- A host must durably claim a due occurrence with its run record before external
  execution, preventing restart from dispatching the same catch-up twice.

This supports a migration with no interval where both old and new schedulers can
run the same definition.

## Commands and synchronization

Adds:

- `listAutomations`;
- `listAutomationTriggerDefinitions`;
- `createAutomation`;
- `updateAutomation`;
- `disposeAutomation`;
- `runAutomation`;
- `fetchAutomationRuns`;
- `previewAutomationSchedule`;
- root catalogue notifications;
- definition/run-summary reducer actions;
- run lifecycle/session/artifact reducer actions;
- client-dispatchable run cancellation.

Presence capabilities use optional objects rather than booleans so they can gain
feature-specific details without changing capability detection.

## Generated surfaces

The contract is propagated through JSON Schema, reference documentation,
TypeScript, Rust, Go, Swift, Kotlin, reducer fixtures, and round-trip fixtures.
The guide documents authority, cron semantics, imports, revisions, lifecycle,
cancellation, retention, reconciliation, and security.

## Deliberate boundaries

- Sharing is within one authority; AHP does not replicate independent hosts.
- Event types remain authority-defined.
- Run-history deletion is not part of this revision.
- Local and managed/cloud execution can be represented by separate authorities.
- Session-list promotion is not modeled here yet; immutable automation
  provenance must not be cleared to change client presentation.

## Main merge

Merged current AHP `main` through `d6bf47d2`, preserving its scoped
customization-enablement model. Additive generated-client conflicts were
resolved by updating the Rust generator and regenerating every client/schema.

## Validation

- `npm run generate`
- `npm run test`
- `cargo test --workspace` under `clients/rust`
- `npm run docs:build`
- 100% reducer branch coverage
- change-fragment and release-metadata verification

## Companion implementation

Companion VS Code draft PR:
https://github.com/microsoft/vscode/pull/330463


types/channels-automation/state.ts
```diff
@@ -0,0 +1,203 @@
...
+  /** IANA time-zone identifier. */
+  timeZone: string;
+}
+
+/** @category Automation State */
+export type AutomationSchedule =
+  | AutomationHourlySchedule
+  | AutomationDailySchedule
+  | AutomationWeeklySchedule
+  | AutomationCronSchedule;
```
connor4312: Cron schedules can express all these things -- do we need to specially represent hours/daily/weekly automations on the protocol?

types/channels-automation/state.ts
```diff
@@ -0,0 +1,203 @@
...
+  time: AutomationLocalTime;
+  /** IANA time-zone identifier. */
+  timeZone: string;
+}
+
+/** @category Automation State */
+export interface AutomationCronSchedule {
+  kind: AutomationScheduleKind.Cron;
+  /** Standard five-field Unix cron expression. */
+  expression: string;
```
connor4312: Various systems have various degrees of supported cron features. We should specify what AHP cron formats support. I suggest something along the lines of the capabilities from https://www.ibm.com/docs/en/db2-as-a-service?topic=task-unix-cron-format (random example I found online)

docs/guide/automations.md
```diff
@@ -0,0 +1,46 @@
+# Automations
+
+Automations let several AHP clients share durable agent-session workflows
```
connor4312: This is quite minimal, we should expand it a bit. See other guide pages for an example.

types/channels-automation/state.ts
```diff
@@ -0,0 +1,203 @@
...
+  nextRunAt?: string;
+  /** Newest-first retained run summaries. */
+  runs: AutomationRunSummary[];
+  runsNextCursor?: string;
+  runtime?: AutomationRuntimeState;
+  operations: AutomationOperation[];
+  createdAt: string;
+  modifiedAt: string;
+  _meta?: Record<string, unknown>;
+}
```
connor4312, line 326: The changes in general lack documentation that we have on other parts of the protocol, which makes it a little harder to review. Can you expand a bit?

types/channels-automation/state.ts
ulugbekna: Addressed in a8643069. The hourly/daily/weekly schedule variants were removed; AutomationSchedule is now one portable five-field cron expression plus an IANA time zone.

types/channels-automation/state.ts
ulugbekna: Addressed in a8643069. The protocol now specifies the exact five fields, numeric and named values, wildcards, ranges, lists, steps, Sunday 0/7 handling, Unix day-of-month/day-of-week matching, and explicitly unsupported seconds/years/macros/Quartz extensions. Schedule capabilities now only add an optional minimum interval.

docs/guide/automations.md
ulugbekna: Addressed in a8643069. The Automations guide now covers capabilities, authority, resource ownership, the full cron grammar, event discovery, revisions, migration, run lifecycle, linked sessions, cancellation, retention, reconciliation, and security.

types/channels-automation/state.ts
ulugbekna: Addressed in a8643069. Added detailed normative docblocks across automation state, actions, commands, capabilities, root notifications, session provenance, and automation-run state/actions; regenerated reference docs, schemas, and all clients.

types/channels-automation/state.ts
```diff
@@ -0,0 +1,374 @@
...
+   * Newest-first retained run summaries. This is a bounded window; use
+   * `fetchAutomationRuns` when {@link runsNextCursor} is present.
+   */
+  runs: AutomationRunSummary[];
+  /** Opaque cursor for the next older run-history page. */
+  runsNextCursor?: string;
+  /** Optional host-resolved execution context. */
+  runtime?: AutomationRuntimeState;
+  /** Operations currently permitted for this automation. */
+  operations: AutomationOperation[];
```
dmitrivMS, line 319: AI Review: `operations` is documented as authoritative and able to change over time, but this channel has no action/reducer path that updates it. A subscribed client can therefore retain stale permissions when run state changes whether update/dispose/run is allowed. Please add an observable replacement action (and consider `runtime` if it is also live), or define another synchronization mechanism.

types/common/messages.ts
```diff
@@ -174,6 +192,14 @@ export interface CommandMap {
   'sessionConfigCompletions': { params: SessionConfigCompletionsParams; result: SessionConfigCompletionsResult };
   'completions': { params: CompletionsParams; result: CompletionsResult };
   'invokeChangesetOperation': { params: InvokeChangesetOperationParams; result: InvokeChangesetOperationResult };
+  'listAutomations': { params: ListAutomationsParams; result: ListAutomationsResult };
+  'listAutomationTriggerDefinitions': { params: ListAutomationTriggerDefinitionsParams; result: ListAutomationTriggerDefinitionsResult };
+  'createAutomation': { params: CreateAutomationParams; result: null };
```
dmitrivMS: AI Review: An import retry with the same `AutomationImportIdentity` may resolve to a previously imported automation independently of the newly requested URI, but this command returns `null`. The caller cannot discover the canonical existing resource in that case. Could `createAutomation` return the resolved automation URI?

types/channels-automation/reducer.ts
```diff
@@ -0,0 +1,73 @@
...
+          return false;
+        }
+        known.add(run.resource);
+        return true;
+      })];
+      const next: AutomationState = { ...state, runs };
+      if (action.nextCursor === undefined) {
+        delete next.runsNextCursor;
+      } else {
+        next.runsNextCursor = action.nextCursor;
```
dmitrivMS: AI Review: `automation/runsLoaded` does not identify the cursor it consumed, yet it unconditionally replaces `runsNextCursor`. If requests for C1 and C2 overlap and their broadcast actions arrive out of order, a late C1â†’C2 page can overwrite C2â†’C3 and move pagination backwards. Please carry the consumed cursor and advance only on a match, or explicitly require serialized loads.

clients/typescript/src/client/hosts/runtime.ts
```diff
@@ -123,6 +126,7 @@ export function snapshotHandle(shared: HostShared): HostHandle {
     protocolVersion: shared.protocolVersion,
     serverSeq: shared.serverSeq,
     defaultDirectory: shared.defaultDirectory,
+    automations: shared.automations,
```
dmitrivMS, line 129: AI Review: `snapshotHandle` exposes `shared.automations` by reference while the API says snapshots should be treated as immutable. `readonly automations` only prevents replacing the property; a consumer can mutate nested capability fields and thereby alter later snapshots/runtime-owned state. Please clone this small capability graph when constructing the snapshot.

clients/typescript/test/client.test.ts
```diff
@@ -113,6 +113,35 @@ test('subscribe attaches before sending the request and fans out an action', asy
     delivery: { maxLatencyMs: 100 },
     view: { turns: 30 },
   });
+
+  test('root automation catalogue notifications reach subscriptions', async () => {
```
dmitrivMS: AI Review: This test is declared inside an already-running async test and its returned promise is not awaited. The parent can finish and cancel/skip these assertions, so the intended notification coverage is unreliable. Please move it to top level (or make it an explicitly awaited subtest).

scripts/generate-json-schema.test.ts
```diff
@@ -173,18 +173,14 @@ describe('generated JSON schemas', () => {
         );
       });
 
-      it('inherits request metadata from BaseParams', () => {
+      it('preserves automation schedule restrictions', () => {
```
dmitrivMS, line 176: AI Review: This replaces the existing `BaseParams._meta` inheritance regression test rather than adding schedule coverage alongside it. The new automation assertion should not remove unrelated schema-generator coverage; please restore the `_meta` test as a separate case.

types/channels-automation-run/state.ts
```diff
@@ -0,0 +1,314 @@
...
+
+/**
+ * Cause recorded for a run created by one of the automation's triggers.
+ *
+ * @category Automation Run State
+ */
+export interface AutomationTriggeredRunCause {
+  kind: AutomationRunCauseKind.Trigger;
+  /** Matches the stable {@link AutomationTrigger.id} in the definition. */
+  triggerId: string;
```
dmitrivMS, line 71: AI Review: Is `triggerId` alone sufficient immutable provenance after the owning definition is revised? A trigger can be removed or its ID reused with different semantics, leaving historical runs ambiguous. Consider recording the automation definition revision (or an immutable trigger snapshot) in the cause.

docs/guide/automations.md
```diff
@@ -0,0 +1,481 @@
...
+`ahp-automation-run:` URI:
+
+```typescript
+{
+  channel: 'ahp-automation:/triage'
+  requestId: 'client-generated-idempotency-key'
+}
+```
+
+`requestId` is durable. Retrying with the same automation and request id
```
dmitrivMS, line 380: AI Review: What retention guarantee does â€œdurableâ€ impose on the `(automation, requestId) â†’ run` mapping? Run history is explicitly bounded, so if the mapping is pruned with the run, a sufficiently late retry can create a duplicate. Please define whether idempotency survives history pruning and, if not indefinitely, its minimum retention window.

types/channels-automation/state.ts
```diff
@@ -0,0 +1,374 @@
...
+ * Operations the host currently permits for an automation.
+ *
+ * The list on {@link AutomationState.operations} is authoritative and may
+ * change over time. Clients MUST NOT infer permission from capabilities alone:
+ * capabilities describe what the host implementation can support, while
+ * operations describe what is allowed for this particular automation now.
+ *
+ * @category Automation State
+ */
+export const enum AutomationOperation {
```
connor4312, line 39: Do we expect these to be the only kinds of actions for automations? If yes, are they actually going to vary by automation or should then just be capabilities at the e.g. host level?

We have `ChangesetOperation` already which is a richer/more flexible type. When read "AutomationOperation" I initially assume that should be following the same pattern.

types/channels-automation/state.ts
```diff
@@ -0,0 +1,374 @@
...
+   * process is unavailable are handled according to the trigger's
+   * {@link AutomationMisfirePolicy}.
+   */
+  HostLifetime = 'hostLifetime',
+  /**
+   * Automatic triggers continue to be evaluated independently of connected
+   * clients and any particular interactive host process.
+   */
+  Managed = 'managed',
+}
```
connor4312: - How does a client use/represent this information?
- Would a single AH want to mix execution lifetimes? (e.g. showing both cloud and local automations in our AH.) In this model a host represents a single lifetime in its capabilities for all automations.

types/channels-automation/state.ts
```diff
@@ -0,0 +1,374 @@
...
+  /**
+   * Selected {@link AutomationTriggerEventDefinition.id | event ids} for this
+   * trigger type.
+   */
+  events: string[];
+  /**
+   * Values described by {@link AutomationTriggerDefinition.configSchema}.
+   * Clients MUST preserve unknown entries when editing other fields.
+   */
+  config?: Record<string, unknown>;
```
connor4312, line 174: Would we want additional information from the AutomationTriggerDefinition/AutomationTriggerEventDefinition to aid in display here? Since looking them up again required a `ListAutomationTriggerDefinitionsParams` method call and potentially some enumeration under the hood...

types/channels-automation/state.ts
```diff
@@ -0,0 +1,374 @@
...
+ */
+export interface AutomationRuntimeState {
+  /**
+   * Effective working directories after host-side preparation, such as
+   * materializing a managed workspace.
+   */
+  workingDirectories?: URI[];
+  /** Opaque host-defined runtime metadata. */
+  _meta?: Record<string, unknown>;
+}
```
connor4312: I'm unclear on how this is used. Automations already have their own `session.workingDirectories`, how is this different?

types/channels-automation/state.ts
```diff
@@ -0,0 +1,374 @@
...
+  /** Current {@link AutomationDefinition.enabled} value. */
+  enabled: boolean;
+  /** Number of automatic triggers in the current definition. */
+  triggerCount: number;
+  /** Earliest schedule occurrence awaiting evaluation, as an ISO 8601 timestamp. It may be in the past while catch-up is pending. */
+  nextRunAt?: string;
+  /** Most recent retained run, when any run exists. */
+  lastRun?: AutomationRunSummary;
+  /** Monotonic definition revision used for optimistic concurrency. */
+  revision: number;
```
connor4312: this does not appear to actually be used

types/channels-automation/state.ts
```diff
@@ -0,0 +1,374 @@
...
+export interface AutomationState {
+  /** URI of this automation channel. */
+  resource: URI;
+  /** Current durable definition. */
+  definition: AutomationDefinition;
+  /**
+   * Monotonically increasing definition revision. Clients pass the revision
+   * they observed as `updateAutomation.expectedRevision`.
+   */
+  revision: number;
```
connor4312: This is not something we should encode in this layer. Actions bear sequence numbers, so if we use actions for this we get optimistic concurrency already (hosts may choose to reject actions if the client would be outside)

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,293 @@
...
+ * @direction Client → Server
+ * @messageType Request
+ * @version 1
+ */
+export interface ListAutomationsParams extends BaseParams, PaginatedParams {
+  /** Automation catalogues are listed from the root channel. */
+  channel: 'ahp-root://';
+  /** Optional exact filter on {@link AutomationDefinition.enabled}. */
+  enabled?: boolean;
+}
```
connor4312: Alternately, we could build it as the `ahp-automation[s]` channel is a catalog of all `AutomationState`. Unlike sessions, automations are not going to be as entirely unbounded to the point that we'd need to page aggressively imo (and we could do something like we did for turns where there is continuation.) That would keep more things in state and we avoid the special "re-run this command after reconnect because root catalogue notifications are not replayed" case

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,293 @@
...
+export interface AutomationImport {
+  /** Stable namespace identifying the source implementation or store. */
+  source: string;
+  /** Identifier shared by every item in one import attempt. */
+  batchId: string;
+  /** Stable source-side identifier for this definition within the batch. */
+  itemId: string;
+  /** Source schedule occurrences to retain until the imported definition is enabled. */
+  triggerNextRuns?: AutomationImportTriggerNextRun[];
+}
```
connor4312: Is this something that should be in the protocol or is this just a legacy path for VS Code to use for migration? (what does batchId/itemId in the protocol?))

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,293 @@
...
+  /** Client-chosen `ahp-automation:` URI for the new definition. */
+  channel: URI;
+  /** Complete initial definition. */
+  definition: AutomationDefinition;
+  /**
+   * Optional legacy import state. When present, {@link definition} MUST be
+   * disabled so automatic triggers cannot run before migration cutover.
+   */
+  import?: AutomationImport;
+}
```
connor4312: Could also go to an action if we are a catalog-state

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,293 @@
...
+  message?: Message;
+  /** Replacement session template. */
+  session?: AutomationSessionTemplate;
+  /** Replacement automatic-trigger enabled state. */
+  enabled?: boolean;
+  /** Complete replacement trigger list. */
+  triggers?: AutomationTrigger[];
+  /** Complete replacement implementation-defined metadata. */
+  _meta?: Record<string, unknown>;
+}
```
connor4312: This should be an action on the automation channel rather than a method

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,293 @@
...
+ * @version 1
+ */
+export interface UpdateAutomationParams extends BaseParams {
+  /** Target `ahp-automation:` URI. */
+  channel: URI;
+  /** Revision on which the client based {@link changes}. */
+  expectedRevision: number;
+  /** Editable fields to replace. */
+  changes: AutomationDefinitionPatch;
+}
```
connor4312: This should be an action on the automation channel rather than a method

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,293 @@
...
+
+/**
+ * Result identifying the existing or newly created run.
+ *
+ * @category Commands
+ */
+export interface RunAutomationResult {
+  /** Subscribable `ahp-automation-run:` URI. */
+  run: URI;
```
connor4312:
```suggestion
  resource: URI;
```

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,293 @@
...
+ * @version 1
+ */
+export interface PreviewAutomationScheduleParams extends BaseParams {
+  /** Schedule preview is requested from the root channel. */
+  channel: 'ahp-root://';
+  /** Portable AHP cron schedule to evaluate. */
+  schedule: AutomationSchedule;
+  /** Requested maximum number of future occurrences; the host MAY cap it. */
+  count?: number;
+}
```
connor4312: I'm unconvinced we need a whole protocol method for this. I would not be opposed to including a test case for cron scheduling in client libraries in this repo (like action tests, we should have a JSON assertion file and then have each language test against that fixture for consistency)

types/channels-automation-run/state.ts
```diff
@@ -0,0 +1,314 @@
...
+   */
+  event?: Record<string, unknown>;
+}
+
+/**
+ * Immutable provenance describing why a run was created.
+ *
+ * @category Automation Run State
+ */
+export type AutomationRunCause =
```
connor4312: We have the notion of `origin`'s already for message and chat provenance, I suggest we use the same terminology here


```suggestion
export type AutomationRunOrigin =
```

types/channels-automation-run/state.ts
```diff
@@ -0,0 +1,314 @@
...
+  | AutomationCompletedRunLifecycle
+  | AutomationFailedRunLifecycle
+  | AutomationCancelledRunLifecycle;
+
+/**
+ * Operations the host currently permits for a run.
+ *
+ * @category Automation Run State
+ */
+export const enum AutomationRunOperation {
```
connor4312: Same Q on operations here -- is this a capability or open ended operation

types/channels-automation-run/state.ts
```diff
@@ -0,0 +1,314 @@
...
+ * Fetchable output produced at run scope rather than by one specific session.
+ *
+ * The inherited {@link ContentRef} identifies how the client obtains the
+ * content. Session-specific edits, transcripts, and tool results remain on
+ * their session and chat channels.
+ *
+ * @category Automation Run State
+ */
+export interface AutomationRunArtifact extends ContentRef {
+  /** Stable artifact id within this run, used by artifact actions. */
```
connor4312: 1. I don't see any notion of artifact actions in this PR
2. These should be unique already by `ContentRef.uri`, no need for another ID

types/channels-automation-run/state.ts
```diff
@@ -0,0 +1,314 @@
...
+ *
+ * @category Automation Run State
+ */
+export type AutomationRunLifecycle =
+  | AutomationPendingRunLifecycle
+  | AutomationRunningRunLifecycle
+  | AutomationBlockedRunLifecycle
+  | AutomationCompletedRunLifecycle
+  | AutomationFailedRunLifecycle
+  | AutomationCancelledRunLifecycle;
```
connor4312, line 191: It seems in these we kind of reinvent/mirror the `SessionInputRequest` / `SessionStatus` but use slightly different terminology/interfaces. Can we consilidate these?

types/channels-automation/state.ts
ulugbekna: We're at early stage of automations, so a bit hard to say if we expect other actions in future. 

> are they actually going to vary by automation or should then just be capabilities at the e.g. host level?

If we in future support automations to come for plugins, I would expect those would need to miss the "update" operation? (or any time when we would like an automation to be un-updatable/un-deletable, eg admin policy-driven automations)

types/channels-automation/state.ts
ulugbekna: Removed this for now until we have a more clear strategy on cloud-run automations.

clients/swift/AgentHostProtocol/Tests/AgentHostProtocolClientTests/MultiHostClientTests.swift
```diff
@@ -132,6 +132,45 @@ final class MultiHostClientTests: XCTestCase {
...
+                mode: mode,
+                automationCapabilities: capabilities
+            )
+        ).withInitialSubscriptions([RootResourceURI])
+        _ = try await multi.add(config)
+        await waitForHostState(multi, id: "local") { $0.isConnected }
+
+        let initialValue = await multi.host("local")
+        let initial = try XCTUnwrap(initialValue)
+        XCTAssertEqual(initial.automations?.execution.lifetime, .managed)
```
dmitrivMS: AI Review: `AutomationCapabilities` has no `execution` or `lifetime` member, so this assertion and the equivalent reconnect assertion on line 167 fail to compile. Remove them or assert capabilities that are actually configured, such as `create` and `runHistoryLimit`.

clients/swift/AgentHostProtocol/Sources/AgentHostProtocol/Reducers.swift
```diff
@@ -915,7 +916,8 @@ public func isClientDispatchable(_ action: StateAction) -> Bool {
          .sessionCustomizationToggled,
          .sessionMcpServerStartRequested, .sessionMcpServerStopRequested,
          .sessionIsReadChanged,
-         .sessionIsArchivedChanged:
+         .sessionIsArchivedChanged,
+         .automationRunCancelRequested:
```
dmitrivMS, line 936: AI Review: The protocol marks `automation/createRequested`, `automation/updateRequested`, and `automation/removed` as client-dispatchable, but Swift recognizes only `automationRun/cancelRequested`. Integrations using these exported validation helpers will reject valid catalogue mutations. Add all three strings and corresponding `StateAction` switch cases, with tests.

clients/typescript/src/client/state-mirror.ts
```diff
@@ -131,5 +167,23 @@ export class AhpStateMirror {
       this.changesetsMap.set(channel, changesetReducer(current, action as ChangesetAction));
       return;
     }
+    if (channel === AUTOMATIONS_URI) {
+      this.setAutomationCatalog(automationReducer(this.automationCatalogState, action as AutomationAction));
```
dmitrivMS, line 171: AI Review: This reduces every server-pushed automation envelope without checking `rejectionReason`. A rejected `automation/removed` therefore deletes the catalogue entry even though authoritative state remains unchanged, violating the documented reconciliation semantics. Skip rejected envelopes before reduction, audit equivalent mirror paths, and add a rejected-removal test.

types/channels-automation/commands.ts
```diff
@@ -0,0 +1,122 @@
...
+ * @messageType Request
+ * @version 1
+ */
+export interface RunAutomationParams extends BaseParams {
+  /** Manual runs are scoped to the catalogue channel. */
+  channel: 'ahp-automations://';
+  /** Target {@link AutomationState.resource}. */
+  automation: URI;
+  /**
+   * Durable client-generated idempotency key. Retrying with the same key and
```
dmitrivMS, line 74: AI Review: The durable key is defined only as `(automation, requestId)`, while `requestId` is an arbitrary client-generated string. Two clients can therefore reuse a value such as `1`, causing one manual run to resolve to the other client's run. Scope the durable key by reconnect-stable `clientId`, or require globally unique request IDs.

types/common/commands.ts
```diff
@@ -266,8 +274,86 @@ export interface InitializeResult {
...
+ *
+ * @category Commands
+ */
+export interface AutomationScheduleCapabilities {
+  /**
+   * Smallest permitted interval between consecutive occurrences produced by
+   * {@link AutomationSchedule.expression}. Omission means no restriction beyond
+   * the cron format's one-minute resolution.
+   */
+  minIntervalMinutes?: number;
```
dmitrivMS, line 343: AI Review: `minIntervalMinutes` is emitted as JSON Schema `number`, so values such as `1.5` validate in TypeScript while generated Go, Rust, Kotlin, and Swift bindings require integers. The same mismatch affects `runHistoryLimit` and `AutomationRunSummary.sessionCount`. Generate integer schemas with minimum 1 for the interval and 0 for the counts, and add cross-language validation tests.

types/channels-automation/state.ts
```diff
@@ -0,0 +1,344 @@
...
+ * capabilities describe what the host implementation can support, while
+ * operations describe what is allowed for this particular automation now.
+ *
+ * @category Automation State
+ */
+export const enum AutomationOperation {
+  /** Replace editable fields using {@link AutomationUpdateRequestedAction | `automation/updateRequested`}. */
+  Update = 'update',
+  /** Permanently remove the automation using {@link AutomationRemovedAction | `automation/removed`}. */
+  Dispose = 'dispose',
```
connor4312: If this action is called remove, the operation should be called remove too

types/channels-automation/state.ts
```diff
@@ -0,0 +1,344 @@
...
+ * {@link AutomationState} rather than in the definition.
+ *
+ * @category Automation State
+ */
+export interface AutomationDefinition {
+  /** Human-readable automation name. */
+  title: string;
+  /**
+   * Initial message sent to every newly created run session. Its
+   * {@link Message.origin} kind MUST be {@link MessageKind.User}.
```
connor4312: I think we might actually have the messages inserted with a distinct origin so that a client can tell it came from an automation, what do you think?

connor4312 (dismissed):

Yoyokrazy (approved):

eleanorjboyd (approved):

PR #397, opened 2026-08-13
session: scoped customization enablement

connor4312:
> Note: I'm putting this up for review tonight, but will refine it more a bit tomorrow -- landing it to avoid conflicts with ongoing vscode dev

## Summary
- Adds scoped `CustomizationEnablement` decisions for plugins and MCP servers, with precedence ordered Session, Workspace, then Global.
- Adds client-published child enablement and MCP client-bundled metadata; directories and leaf customizations retain their plain `enabled` semantics.
- Ports reducer behavior and shared fixtures across all generated clients.

## Wire contract
Producers publish `enablement` entries in descending specificity. Consumers may use `enablement?.[0]?.enabled ?? true`; absent or empty enablement means enabled. The host emits at most one Workspace decision for the session primary working directory.

Consumer: microsoft/vscode#330566

### DIFF
--- a/scripts/generate-go.ts
+++ b/scripts/generate-go.ts
@@ -695,7 +695,7 @@ const STATE_ENUMS = [
   'ToolCallRiskAssessmentStatus',
   'ToolCallCancellationReason',
   'ConfirmationOptionKind', 'ToolCallContributorKind',
-  'ToolResultContentType', 'CustomizationType', 'CustomizationLoadStatus', 'TerminalClaimKind',
+  'ToolResultContentType', 'CustomizationType', 'CustomizationEnablementKind', 'CustomizationLoadStatus', 'TerminalClaimKind',
   'McpServerStatus', 'McpAuthRequiredReason',
   'ChangesetStatus', 'ChangesetOperationStatus', 'ChangesetOperationScope', 'ResourceChangeType',
 ];
@@ -1315,6 +1315,11 @@ function generateStateFile(project: Project): string {
     }
   }
 
+  lines.push('// ─── Customization Enablement Union ───────────────────────────────────────');
+  lines.push('');
+  lines.push(generateCustomizationEnablementGo());
+  lines.push('');
+
   lines.push(generateToolInput());
   lines.push('');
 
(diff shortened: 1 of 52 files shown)


types/channels-session/reducer.ts
```diff
@@ -105,6 +108,27 @@ function updateMcpServerCustomization(
...
+ * retain their legacy `enabled` field, derived from the incoming decisions.
+ */
+function applyCustomizationEnablement(customization: Customization, enablement: readonly CustomizationEnablement[]): Customization;
+function applyCustomizationEnablement(customization: ChildCustomization, enablement: readonly CustomizationEnablement[]): ChildCustomization;
+function applyCustomizationEnablement(customization: Customization | ChildCustomization, enablement: readonly CustomizationEnablement[]): Customization | ChildCustomization {
+  switch (customization.type) {
+    case CustomizationType.Plugin:
+    case CustomizationType.McpServer: {
+      if (enablement.length > 0) {
+        return { ...customization, enablement: [...enablement] };
```
dmitrivMS, line 122: AI Review: `session/customizationToggled` is client-dispatchable and schema-valid arrays can arrive in any order, but this reducer persists caller order while consumers derive the effective value from element 0. The wire contract says producers MUST sort, yet the server-validation table has no rule for rejecting unsorted or duplicate-scope input, so a malformed client action can be optimistically applied and echoed with the wrong effective state. Please either canonicalize/deduplicate at this boundary in every reducer, or specify server rejection semantics and add shared fixtures for unsorted and duplicate kinds.

scripts/generate-go.ts
```diff
@@ -1610,6 +1615,74 @@ const CHAT_SOURCE_UNION: UnionConfig = {
...
+\t\t}
+\t\te.Value = &value
+\tcase "session":
+\t\tvar value CustomizationEnablementSession
+\t\tif err := json.Unmarshal(data, &value); err != nil {
+\t\t\treturn err
+\t\t}
+\t\te.Value = &value
+\tdefault:
+\t\treturn &json.UnmarshalTypeError{Value: "CustomizationEnablement"}
```
dmitrivMS, line 1673: AI Review: Is `CustomizationEnablement` intentionally a closed union? Go, Kotlin, and Swift all fail the containing payload on an unknown `kind`, while AHP clients are released independently and several state unions preserve unknown variants. If another scope may be added within a compatible protocol version, this needs an unknown-preserving branch plus a round-trip fixture; otherwise, please document that adding a scope requires an incompatible protocol version.

dmitrivMS (commented):
AI Review: Council review identified two additional protocol contract questions.

types/channels-session/state.ts
```diff
@@ -828,6 +845,17 @@ export interface PluginCustomization extends ContainerCustomizationBase {
...
+   * Explicit enablement decisions for children this plugin contributes,
+   * keyed by child name (for MCP servers, the server name as it appears in
+   * the bundled `.mcp.json`).
+   *
+   * Bundled children are discovered by the host rather than published by the
+   * client, so the client cannot attach `enablement` to them directly. This
+   * carries the client's global decision for each one; the host applies it
+   * under the child's durable key.
+   */
+  childEnablement?: Record<string, CustomizationEnablement[]>;
```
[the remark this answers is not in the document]
dmitrivMS: AI Review: The council independently confirmed this contract mismatch. The array shape also permits zero or multiple Global entries, even though the field is described as carrying the client's single Global decision for each child. A single Global-only decision value would make both invalid scope and invalid cardinality unrepresentable.

alexr00 (approved):

PR #402, opened 2026-08-14
session: add atomic working directory replacement

### LINKED ISSUE #392: workingDirectories collapsed session cwd into the tool-access set, leaving cwd with no mutation path
## Summary

`0.7.0` replaced the scalar `SessionState.workingDirectory` with the `workingDirectories` set, and in doing so collapsed two different concepts into one field:

- **where the session's agent is working** (its cwd — what a `cd` command changes), and
- **which roots the agent's tools may touch** (a multi-root access set).

The replacement field is documented with the *access* meaning, and the two mutation actions it introduced (`session/workingDirectorySet` / `session/workingDirectoryRemoved`) are membership operations. The result is that the cwd concept survived the migration with **no mutation path**, and the spec now contradicts itself about whether the field is identity or mutable state.

## Evidence

**0.6** — scalar, documented as cwd:

> The default working directory URI for this session.

**0.7** — set, documented as access:

> The working directories the session's agent has tool **access** to, as maintained by the `session/workingDirectorySet` / `session/workingDirectoryRemoved` actions.

**The actions are membership-only.** `SessionWorkingDirectorySetAction`:

> A working directory was **added** to the session's `SessionState.workingDirectories` set. Membership semantics keyed by the directory URI: the reducer **appends** `directory` when the set does not already contain it (creating the set if absent) and is a no-op when it is already present.

There is no reorder action and no replace action, on `0.7.0` or on `main`. So "move this session's cwd" is expressible only as remove-entry-0 + append + reorder, of which the first is forbidden under `immutablePrimary` and the third does not exist.

**`immutablePrimary` cannot describe a movable primary.** It is binary:

> `true`: the first working directory is an immutable primary: it is fixed for the lifetime of the session — clients MUST NOT remove or reorder it. […] When absent or `false`, all directories are equal peers and any of them may be removed.

`true` denies the primary can move; `false` denies there is a primary at all. A host whose agent has a real, movable cwd plus additional roots has no honest way to advertise itself.

**The spec disagrees with itself about mutability.** `docs/specification/session-channel.md` lists `workingDirectories` under session-wide **identity**, alongside `resource` / `provider` / `createdAt`. `docs/guide/state-model.md` lists the working directory among the **mutable** fields streamed via `root/sessionSummaryChanged` — and `PartialSessionSummary.workingDirectories` exists, so the catalog half *can* carry a change that the session-state half has no action to apply.

Compare `title`, which got both halves: `session/titleChanged` **and** the summary notification. The working directory has only the summary half.

## A factual correction

`MultipleWorkingDirectoriesCapability.immutablePrimary` justifies itself with:

> Advertised by backends whose agent process is rooted at a single directory that cannot change once the session has started (e.g. **the SDK's primary `workingDirectory`**).

For the Copilot SDK this is not accurate. Its `session.metadata.setWorkingDirectory` documents:

> it does **NOT** change the process working directory (**a session's cwd is per-session, not process-global**). For local sessions the runtime validates the target first (an absolute path that exists on disk) and re-bases the permission primary directory; a rejected validation fails the call before anything is mutated, persisted, or emitted.

It is validated-then-applied atomically per session, and has a passing end-to-end test exercising it mid-session. Since that example is doing load-bearing work in the capability's rationale, it is worth correcting regardless of the outcome below.

## Concrete impact

An agent host backed by that SDK cannot expose a working-directory change to AHP clients at all, even though the backend supports it and validates it atomically. Clients that want the equivalent of `cd` on an existing session must either fall back to a vendor `extensions/` method (which desyncs `SessionState` from `SessionSummary`, since only the latter can carry the change) or tell users to create a new session in the directory they wanted.

## Proposal

Add a full-replacement action:

```
session/workingDirectoriesChanged { directories: URI[] }
```

Rationale:

- **Ordered**, so it expresses a primary move without a reorder primitive, and remains meaningful under `immutablePrimary` (the host can still refuse any change to entry `0`).
- **Atomic**, so subscribers never observe the transient intermediate set that a remove-then-append pair produces.
- **Consistent with existing full-replacement actions** such as `session/changesetsChanged` and `session/metaChanged` (the latter explicitly "Replaces `state._meta` entirely").
- It subsumes replace, reorder, and bulk edits without adding three actions.

A host that cannot apply a given change still declines it, exactly as today.

Alternatively, if the primary should stay pinned, a capability that can express "there is a primary and it is movable" plus a `session/primaryWorkingDirectorySet` would also close the gap — but the full-replacement action seems the smaller and more general addition.

Happy to open a PR for whichever shape you prefer.

connor4312:
## Summary

- add client-dispatchable `session/workingDirectoryReplaced` to atomically replace any matching working-directory entry in place
- use the existing URI as a compare-and-swap guard, deduplicate the replacement URI wherever it already exists, and preserve unrelated directory ordering
- add `MultipleWorkingDirectoriesCapability.primaryReplacement` to indicate that the protected index-0 entry may also be replaced
- propagate the action through schemas, generated clients, and the hand-written Go, Rust, Kotlin, and Swift reducers

Replacing an ordinary entry requires `multipleWorkingDirectories`. Targeting index 0 additionally requires `primaryReplacement: true`; `immutablePrimary: true` and `primaryReplacement: true` cannot both be advertised. This keeps `workingDirectories` as the protocol state instead of introducing a separate primary-working-directory field.

## Validation

- `npm run generate`
- `npm test`
- `cd clients/go && go test ./...`
- `cd clients/rust && cargo test -p ahp`
- `cd clients/swift/AgentHostProtocol && swift test --filter FixtureDrivenReducerTests`
- Kotlin `./gradlew test --no-daemon --console=plain` in the prescribed `eclipse-temurin:17-jdk` Podman container

Fixes #392

### DIFF
--- a/clients/go/ahp/reducers.go
+++ b/clients/go/ahp/reducers.go
@@ -925,13 +925,21 @@ func ApplyActionToSession(state *ahptypes.SessionState, action ahptypes.StateAct
 		}
 		return ReduceOutcomeNoOp
 	case *ahptypes.SessionWorkingDirectoryReplacedAction:
-		if len(state.WorkingDirectories) == 0 || state.WorkingDirectories[0] != a.Directory {
+		idx := -1
+		for i, directory := range state.WorkingDirectories {
+			if directory == a.Directory {
+				idx = i
+				break
+			}
+		}
+		if idx < 0 {
 			return ReduceOutcomeNoOp
 		}
 		updated := make([]ahptypes.URI, 0, len(state.WorkingDirectories))
-		updated = append(updated, a.Replacement)
-		for _, directory := range state.WorkingDirectories[1:] {
-			if directory != a.Replacement {
+		for i, directory := range state.WorkingDirectories {
+			if i == idx {
+				updated = append(updated, a.Replacement)
+			} else if directory != a.Replacement {
 				updated = append(updated, directory)
 			}
 		}
--- a/clients/go/ahp/reducers.go
+++ b/clients/go/ahp/reducers.go
@@ -935,6 +935,12 @@ func ApplyActionToSession(state *ahptypes.SessionState, action ahptypes.StateAct
 		if idx < 0 {
 			return ReduceOutcomeNoOp
 		}
+		for i := 0; i < idx; i++ {
+			if state.WorkingDirectories[i] == a.Replacement {
+				state.WorkingDirectories = append(state.WorkingDirectories[:idx], state.WorkingDirectories[idx+1:]...)
+				return ReduceOutcomeApplied
+			}
+		}
 		updated := make([]ahptypes.URI, 0, len(state.WorkingDirectories))
 		for i, directory := range state.WorkingDirectories {
 			if i == idx {
--- a/clients/go/ahp/reducers.go
+++ b/clients/go/ahp/reducers.go
@@ -924,6 +924,19 @@ func ApplyActionToSession(state *ahptypes.SessionState, action ahptypes.StateAct
 			}
 		}
 		return ReduceOutcomeNoOp
+	case *ahptypes.SessionWorkingDirectoryReplacedAction:
+		if len(state.WorkingDirectories) == 0 || state.WorkingDirectories[0] != a.Directory {
+			return ReduceOutcomeNoOp
+		}
+		updated := make([]ahptypes.URI, 0, len(state.WorkingDirectories))
+		updated = append(updated, a.Replacement)
+		for _, directory := range state.WorkingDirectories[1:] {
+			if directory != a.Replacement {
+				updated = append(updated, directory)
+			}
+		}
+		state.WorkingDirectories = updated
+		return ReduceOutcomeApplied
 	case *ahptypes.SessionInputNeededSetAction:
 		id, ok := sessionInputRequestID(a.Request)
 		if !ok {
--- a/clients/go/ahptypes/actions.generated.go
+++ b/clients/go/ahptypes/actions.generated.go
@@ -921,10 +921,12 @@ type SessionWorkingDirectoryRemovedAction struct {
 //
 // This is a targeted compare-and-swap: the reducer is a no-op when
 // {@link SessionState.workingDirectories} does not contain `directory`.
-// Otherwise it replaces that entry in place with `replacement` and removes
-// any other occurrence of `replacement`, preserving every other directory's
-// relative order. For example, `[A, B, C]` with `B → D` becomes `[A, D, C]`,
-// while `[A, B, C]` with `B → C` becomes `[A, C]`.
+// Otherwise it replaces that entry with `replacement` and deduplicates the
+// result, preserving every other directory's relative order. When
+// `replacement` occurs after the target, it moves to the target's position;
+// for example, `[A, B, C]` with `B → C` becomes `[A, C]`. When it occurs
+// before the target, it retains its earlier position and the target is removed;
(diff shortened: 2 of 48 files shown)


examon:
Thanks for turning this around so fast. The compare-and-swap is a better shape than the full-replacement action I proposed in #392 - preserving concurrent peer edits beats clobbering the array, and requiring the backend side effect before the broadcast answers the exact objection copilotd records in its own source for why it rejects the existing mutations today. Not asking for my original shape back.

Two things I think are worth fixing while this is still a draft.

**1. Dedupe can move the protected primary without ever targeting it.**

Fixture `270-session-workingdirectoryreplaced-deduplicates-earlier-replacement.json` makes this normative: `[repo-a, repo-b, repo-c]` with `{directory: repo-c, replacement: repo-a}` becomes `[repo-b, repo-a]`.

`repo-a` was the primary at index 0. Afterwards `repo-b` is. The action targets index 2, so the `primaryReplacement` validation - which only fires when the target is index 0 - never runs. Under either protected-primary mode a client can therefore change the primary through an action that looks entirely unremarkable, which is the one thing the capability exists to prevent.

Narrow fix: when `replacement` already exists at an index *earlier* than the target, keep it at that earlier position and drop the target entry instead, so `[A,B,C]` with `C -> A` gives `[A,B]`. The later-dedupe case in the docs (`[A,B,C]` with `B -> C` giving `[A,C]`) is unaffected, and index 0 becomes unreachable except by targeting it.

**2. The capability flip fails open for clients on the previous version.**

copilotd advertises `immutablePrimary: true` today. To adopt this it sets that false and `primaryReplacement: true`. A 0.7 client ignores the unknown field, sees `immutablePrimary` absent, and applies the released 0.7 rule that all directories are equal peers and any may be removed. A compliant 0.8 host still rejects the removal so state is safe, but the old client's model has silently loosened at the moment index 0 became *more* special.

We are a concrete instance of this: the Copilot CLI shipped its 0.7.0 client this week, and it hides `/cd`, `/worktree` and `/move` and refuses cwd mutation outright specifically because it reads `immutablePrimary`. On the day copilotd flips the flag, that client starts believing index 0 is an ordinary removable peer.

Would you consider allowing both flags true, defined as "generic removal and reorder forbidden, atomic replacement permitted"? Old clients stay conservative, new clients get the operation. A separate enum does not solve it, since an old client ignores an unknown enum value too. As written the exclusion is prose-only - no `oneOf`/`not` in `state.schema.json`, and every generated client emits two independent optional booleans - so a host that sets both produces contradictory obligations with no defined precedence.

**One request.** Could `primaryReplacement` state normatively that index 0 is the session's default cwd and relative-path base? The capability prose calls it "cwd-bearing", but the state model still describes these as equal-peer tool-access roots and the multiroot proposal says the protocol has no other notion of a main directory. Anyone building `cd` on this needs to know that `pwd` and relative tool paths resolve against index 0. It holds for copilotd today, but it is not something the spec lets a client rely on.

Two smaller ones:

- There is no `chat/workingDirectoryReplaced`, and chat subsets must be a subset of the session set. Session `[A,B]` with chat subset `[A]`, then `A -> C`, leaves that chat referencing a directory no longer in the session set. Worth stating the policy even if it is "hosts MUST reject while a subset references the target".
- Nothing says how an accepted replacement orders against in-flight work. Something like "work beginning after an accepted replacement's server sequence MUST resolve against the replacement; a host that cannot serialize replacement against active work MUST reject it" would stop a single turn running some tools under the old root and some under the new.

Happy to write the copilotd side once the shape settles.

osortega (approved):

Issue #268, opened 2026-06-26
Proposal: host-advertised encryption keys for sealing pushed values end-to-end (first application: the `authenticate` token)


colbylwilliams:
### Summary

AHP delivers the `authenticate` bearer token as **cleartext inside the JSON-RPC frame**. The protocol is transport-agnostic and explicitly treats endpoint access as a transport-layer concern ([Transport](docs/specification/transport.md)), so a conformant deployment MAY carry AHP frames over a **shared or brokered transport** — for example a managed pub/sub relay or message broker that terminates TLS and forwards frames. In that topology a sensitive value in a frame is exposed to the intermediary at the AHP layer, even though each individual hop is TLS-protected. AHP has **no protocol-native way for a party to seal a value end-to-end to the host** that ultimately terminates it.

This proposes a vendor-neutral primitive — (1) a host **advertises a public encryption key**, and (2) a counterparty delivers a value **encrypted and self-describing** to that key — and wires it for the most strongly-motivated first application: the **`authenticate` token**.

The token is the sharpest case, and the right place to start, because it is the one value AHP *forces* into cleartext: it rides in a bare `token: string` with **no `_meta` slot** to carry an alternate encoding, where most other carriers already have a `_meta` escape hatch. But the protocol-shaped need — advertise a host key; let a pushed value be encrypted and self-describing — is not specific to the token, so the additions below are deliberately framed to extend to other pushed values **without a redesign**, with the token as the first and only wired application.

### Current model

Token delivery ([Authentication](docs/specification/authentication.md)) pushes a raw Bearer string:

```ts
export interface AuthenticateParams extends BaseParams {
  channel: 'ahp-root://';
  resource: string;        // MUST match AgentInfo.protectedResources
  token: string;           // raw Bearer token — the spec example uses "gho_xxxx"
}
```

- `token` is a **bare string** with no companion field to signal an alternate encoding, an algorithm, or a key id.
- `InitializeResult` has no vendor/extension slot, and there is **no host-level encryption-key field** anywhere in the handshake or root state.
- `ProtectedResourceMetadata` follows RFC 9728 and already carries JOSE vocabulary — `jwks_uri`, `resource_signing_alg_values_supported`, `resource_encryption_alg_values_supported`, `resource_encryption_enc_values_supported` — but per RFC 9728 these describe the **resource server's** capabilities for signing/encrypting *its own responses*, not a key the **agent host** holds for a counterparty to encrypt a pushed value to.

[...]

Issue #221, opened 2026-06-15
authenticate / auth-required can't model per-session, request-correlated OAuth (e.g. MCP server OAuth)


colbylwilliams:
### Summary

AHP's authentication flow (`auth/required` + the `authenticate` command) models *agent-level, statically-declared* protected resources via RFC 6750 bearer-token push. It can't model a *per-session, request-correlated* OAuth flow — the shape MCP servers use — so a host bridging such a flow to AHP clients has no surface to express it.

### Current model

- `auth/required` (server → client) carries a `channel` and a `resource` (an OAuth-protected resource identifier per RFC 9728).
- `authenticate` (client → server) pushes a **Bearer token** for that `resource`. `AuthenticateParams.channel` is narrowed to `'ahp-root://'`, and `resource` MUST match a `ProtectedResourceMetadata.resource` declared in `AgentInfo.protectedResources`:

```ts
export interface AuthenticateParams extends BaseParams {
  channel: 'ahp-root://';
  resource: string; // must match AgentInfo.protectedResources
  token: string;    // bearer token the client already obtained
}
```

So the model is "push me a bearer token for this declared, agent-level resource."

### The gap

This can't represent an OAuth flow that is **per-session** and **correlated to a specific pending request**. As a concrete, publicly-available example, the github/copilot-sdk runtime emits an `mcp.oauth_required` event:

- it carries an `oauthRequestId` ("used to respond via `session.respondToMcpOAuth()`"), the MCP server name/URL, and an optional static client config (grant type, etc.);
- the client is expected to respond **correlated by that request id**, potentially completing an interactive authorization-code grant rather than just handing back a pre-obtained bearer token.

A host bridging such a backend to AHP clients has nowhere to:

1. **advertise the pending request on the session channel** — `auth/required` is geared to root-channel, agent-declared resources, and `authenticate.channel` is `'ahp-root://'`-only; and
2. **receive a request-correlated completion** — `authenticate` carries only `{ resource, token }`, with no field to echo back a specific request/correlation id, and assumes the client already holds a bearer token rather than completing a pending authorization tied to a particular request.

### Suggested direction

Consider extending the authentication surface to cover per-session, request-correlated OAuth — for example:

- a session-channel `auth/required` variant carrying an opaque request/correlation id (and optional client-config hints), and
- an `authenticate` (or sibling) command variant that echoes that id and can carry an authorization result, not only a pre-obtained bearer token.

[...]

Issue #153, opened 2026-05-27
Reconsider host-level agent auth state


Qusic:
Currently each client must independently obtain OAuth tokens and push them to the host via `authenticate`. The token lifecycle (refresh, expiry) is also per-client. In practice this means the host holds N copies of the same Copilot
 token for N connected clients — all granting access to the same GitHub account on the same machine.

   The per-client isolation this model assumes doesn't exist in practice. In the VS Code agent host, all clients share a single process. Any connected client can execute arbitrary commands and access credentials on the host via tool
 calls. Per-client token isolation would only be meaningful with per-client sandboxing, which the current architecture doesn't provide.

   **Suggestion:** Make agent auth state host-level. The host manages tokens for agents like Copilot centrally — obtaining, storing, and refreshing them. Clients authenticate their connection to the host (`--connection-token`), not
 separately with each agent's upstream service.

   Clients could still assist with the initial auth flow when the host has no token (e.g. presenting a device code UI and passing the result back), but the token would be stored and refreshed by the host, shared across all
 connections. This also simplifies reconnection — clients don't need to re-push tokens after a dropped connection.

   [ahpx](https://github.com/TylerLeonhardt/ahpx) already works around this by resolving GitHub tokens from the environment (`GITHUB_TOKEN` / `gh auth token`) client-side — effectively treating agent auth as a host-level concern.

   ### Spec references

   - Per-client auth flow: [authentication.md#L7-L25](https://github.com/microsoft/agent-host-protocol/blob/e49acc65e411c0e93f1633e53fba03ba0fd7f825/docs/specification/authentication.md?plain=1#L7-L25)
   - Per-resource, not per-connection: [authentication.md#L197-L202](https://github.com/microsoft/agent-host-protocol/blob/e49acc65e411c0e93f1633e53fba03ba0fd7f825/docs/specification/authentication.md?plain=1#L197-L202)
   - Per-connection rationale: [authentication.md#L204-L206](https://github.com/microsoft/agent-host-protocol/blob/e49acc65e411c0e93f1633e53fba03ba0fd7f825/docs/specification/authentication.md?plain=1#L204-L206)

connor4312:
[...]

PR #404, opened 2026-08-18
docs: host-owned authentication proposal

TylerLeonhardt:
Proposes an **opt-in** mode in which an agent host holds its own credentials, instead of relying on a client to broker them. Docs-only — no `types/` changes, so no changelog fragment.

An HTML explainer of the same design is attached below.

## Why

Today the client fetches the token and pushes it with `authenticate`. That assumes a client is present and can do OAuth, which breaks for scheduled work, headless runners, and thin clients (a phone, a browser tab, a CI runner). It also means five clients on one machine push five tokens for the same account — #153.

The reverse case is real too: a sandboxed host *should* be held at arm's length and handed only short-lived tokens. So this isn't "host-owned is better" — it's "AHP can only express one of the two, and it needs both."

## Shape

- **`ahp-accounts://`** — a singleton channel carrying what the host *has* (`accounts`), what it *needs* (`challenges`), and what's *in flight* (`pendingAuth`). Never a token.
- **Two commands** — `authBegin` / `authComplete`. Both return data that must not enter state; for `authComplete` that's a security requirement, since actions are echoed to every subscriber and an action carrying the callback would broadcast a live authorization code.
- **Seven actions** — `Set`/`Updated`/`Removed` across the three keyed collections.
- **Six user flows**, walked through in §5.

`authenticate` is untouched and remains the default. A host that declines the capability is fully conformant — [doctrine](https://microsoft.github.io/agent-host-protocol/guide/doctrine) names "a required credential flow" as an explicit anti-goal.

## Things reviewers may want to push on

- **`HostAccount` is an identity only** — no resource, scopes, expiry, renewability, or status. Credentials aren't modelled at all (§7, §9).
- **`AccountConsumer` is asymmetric** — the agent variant carries a required `resource`, the MCP one doesn't. This mirrors an asymmetry AHP already has: `AgentInfo.protectedResources` and `AuthRequiredErrorData.resources` are arrays, while `McpAuthRequirement.resource` is a single object.
- **`authBegin` takes no `scopes`** — the correct request is *what the host already holds* plus *what's being demanded now*, and the client can't know the first half (§8.6).
- **Scopes accumulate and are shared** — a step-up driven by one consumer widens the credential the others use. Called out in §8.4 rather than left implicit.
- **The MCP consumer key is `(uri, name)`** — AHP has no stable cross-session identifier for an MCP server today. Four weaknesses documented in §7.2 rather than papered over.

## If you were at the whiteboard

Everything from the board survives. Six things changed shape, and the reasoning is in §9:

| On the board | In the proposal | Why |
| --- | --- | --- |
| `getUrlToOpen()` / `finishAuth()` | `authBegin` / `authComplete` with an `attemptId` | Two calls with nothing linking them can't survive two clients signing in at once, a client crashing between them, or a host holding two live PKCE verifiers. Same field #221 asks for. |
| Walk every flow until one works | Host advertises at `initialize`, client picks per attempt | Each failed try burns a real browser window or a live device code, and "until one works" can't tell *unsupported* from *cancelled*. |
| Device code = second priority | Device code = the baseline | The motivating case is absent or remote clients. Auth-code-with-loopback assumes the client runs an HTTP server *and* sits beside a browser. |
| "RT" drawn on the board | Nothing in the protocol — host policy | Whether to request a refresh credential is decided by whoever stores it. The user's control is the consent screen. Three attempts to give AHP an opinion here all failed (§9). |
| "Account status channel" | `ahp-accounts://`, not a root-state field | Seeing the account list should be refusable separately from seeing root state. |
| One account per resource | N accounts + a consumer binding | Without a binding the only semantics is "last signed-in wins" — which is the #153 bug. |

## Where your earlier review landed

All five points applied:

1. **"No reason for a client to ask for flows in `initialize`."** Dropped the client capability. `InitializeResult.hostAuth.flows` is the menu; the client names what it will drive *per attempt* in `authBegin`, which is where the decision actually matters.
2. **"Should `accounts` be a map?"** Kept an array — every keyed collection in AHP (`chats`, `customizations`, `activeClients`, `files`, `annotations`, `agents`, `terminals`) is one, and the `Set`/`Updated`/`Removed` convention is written for arrays. A map would be the only one of its kind.
3. **"Why do `id` and `resource` need to be different things?"** They no longer coexist on the account. `HostAccount` is an identity and carries no `resource` at all; `resource` names *what is being asked for*, `accountId` names *which account answers*.
4. **"Should `HostAccount` own `AccountConsumer[]`?"** Yes — adopted exactly. A binding to a non-existent account is now unrepresentable, and removal is atomic.
5. **"We usually prefix like `auth*`."** Renamed to `authBegin` / `authComplete`.

Two of those cascaded further than the original note. Dropping the account's `resource` (3) exposed that `HostAccount` was conflating identity with credential, which ended with credentials not being modelled at all. And `AccountConsumer` (4) turned out to need a `resource` of its own on the agent variant — one Copilot can hold a GitHub account and an Entra account at once.

## Relates

- Subsumes #153 (N copies of one credential) and #221 (request correlation — `attemptId` + `challengeId`)
- Overlaps #268 (sealed tokens); §10 argues the two should be reconciled rather than designed in parallel

## Open questions (§10)

`CustomizationBase.stableId` · a URI normalization rule for `(uri, name)` · whether `McpAuthRequiredReason` should be renamed once agent providers raise challenges too.

### DIFF
--- a/docs/proposals/host-owned-auth.md
+++ b/docs/proposals/host-owned-auth.md
@@ -163,13 +163,13 @@ missing a credential. On the wire they're identical.
 ```
 accounts     accounts the host holds, and what uses each one
 challenges   credentials the host needs and doesn't have   ← flows 1, 2, 3
-pendingAuth  sign-ins happening right now                  ← flows 1–4
+attempts     sign-ins under way, and how each one ended    ← flows 1–4
 ```
 
 A **challenge** is a *need*. An **attempt** is a *try*. One need can outlive
 several tries, so they're separate collections. Flow 1 over time:
 
-| | `challenges` | `pendingAuth` | The user sees |
+| | `challenges` | `attempts` | The user sees |
 | --- | --- | --- | --- |
 | Host starts, no GitHub credential | `[github]` | — | "Sign in to GitHub" |
 | User clicks; device code issued | `[github]` | `[#1 · ABCD-1234]` | "Enter ABCD-1234" |
@@ -205,7 +205,7 @@ sequenceDiagram
     H->>C: action accounts/challengeSet
     C->>U: renders the prompt for that challenge
     U->>C: approves
-    C->>H: authBegin({ target: { challengeId }, flows, redirectUri? })
+    C->>H: authBegin({ target: { challengeId }, flows })
     H-->>C: authorizationUri, or a device code
     C->>U: opens the browser / displays the code
     U->>AS: authenticates and consents
@@ -221,6 +221,19 @@ subscribed to `ahp-accounts://` then has the whole challenge sitting in
 `AccountsState.challenges`, and passes its `id` back at step 5. No lookup, no
 second source — a client only ever echoes a value the host gave it. That's what
 makes it safe for the challenge to carry full `ProtectedResourceMetadata` (§7).
+
+**Nothing here replaces `AuthRequired` (`-32007`).** It stays exactly as it is,
+along with `McpServerAuthRequiredState` and `ToolCallAuthRequiredState`. A
+client that never subscribes to `ahp-accounts://` sees precisely what it sees
+today, in the same places, and keeps brokering tokens with `authenticate` — that
+is the whole point of the feature being additive.
+
+What subscribing buys is the *earlier* and *wider* view: a need visible before
+anything has failed (flow 1), deduplicated across sessions, and the
+`challengeId` that lets the host run the flow instead of the client. A client
+that wants none of that is unaffected, which is why
+`McpAuthRequirement.challengeId` is optional — absent when the host has no
+challenge published for that connection, and simply ignorable when it isn't.
 
 Challenges cover *static* requirements too, not only failures. That's what makes
 flow 1 work: on first run nothing has failed, the host just knows an agent needs
@@ -377,7 +390,7 @@ every flow until one works. Rejected: each failed try can burn a real browser
 window or a live device code, and "until one works" can't tell *flow unsupported*
 from *user cancelled*. Once you can tell those apart, you can work out the overlap
 up front anyway. So the host advertises its flows in
-`InitializeResult.hostAuth`, the client names the ones it will drive in
+`InitializeResult.authentication`, the client names the ones it will drive in
 `authBegin` and is told which got picked, and it can call again with a narrower
 set. The client drives retries; the host never fires blind.
 
@@ -492,81 +505,38 @@ that is what an account-management UI renders; deriving it by walking every
 session's customizations is both expensive and racy. Per-session overrides and
 `usedBy` rollups stay deferred (§10).
 
-### 7.2 The consumer key: a real gap in AHP today
+### 7.2 The consumer key
 
 A binding has to survive a restart, so it needs an id for "this MCP server" that
-means the same thing in every session. **AHP doesn't guarantee one.**
-
-| Candidate | Why it fails |
-| --- | --- |
-| `CustomizationBase.id` | Documented as a *"session-unique opaque identifier… minted by whoever publishes the customization."* Unique **within** a session; carries no promise **across** sessions. |
-| `uri` alone | For inline declarations, `uri` points to the **containing file** — several MCP servers in one `plugins.json` share it. |
-| `range` | Narrows within the file, but shifts the moment anyone edits it. |
-| `name` alone | Human-readable, with no documented uniqueness guarantee. |
-
-The failure is worse than a lost preference. Take a host that mints ids as
-counters in discovery order — allowed by "opaque" — and let discovery order change
-because a server was added, disabled, or enumerated by a racing filesystem walk:
-
-```
-Session 1   [ github-work (c1), github-personal (c2) ]   binding: c1 → work account
-Session 2   [ github-personal (c1), github-work (c2) ]   binding: c1 → work account
-            ⇒ the personal server now authenticates as the work account
-```
-
-A host minting per-session GUIDs gets the harmless version: the binding matches
-nothing and quietly falls back to "last signed-in wins." **Both hosts are
-conformant**, so the protocol currently permits the credential-misbinding one.
-That's the exact boundary multi-account exists to enforce, so it can't be left to
-luck.
+means the same thing in every session. AHP has no such id today:
+`CustomizationBase.id` is documented as *session-unique*, `uri` alone collides
+for servers declared inline in one manifest, `range` shifts on edit, and `name`
+alone has no uniqueness guarantee.
 
 **This proposal keys on `(uri, name)`** — the manifest, plus the declaration's
-name inside it. It survives file edits, tells inline siblings apart, and needs no
-change to any existing type, so it can't stall on someone else's decision.
+name inside it. It needs no change to any existing type, so it can't stall on
+someone else's decision. It does need one rule the spec doesn't state, so this
+proposal adds it: a host **MUST NOT** publish two MCP-server customizations
+sharing a `(uri, name)` in one session. In practice that already holds — an
+`mcpServers` block is a map keyed by name.
 
-It does need one rule the spec doesn't state today, so this proposal adds it: a
-host **MUST NOT** publish two MCP-server customizations sharing a `(uri, name)`
-in one session. In practice this already holds — an `mcpServers` block is a map
-keyed by name — but the key can't rest on a convention that lives outside the
-protocol.
+Mostly this is a host implementation concern, and hosts already have to solve it
+to keep grants across restarts. Two things are worth pinning in the spec rather
+than leaving to each host:
 
-#### Four ways it breaks
+- **URI comparison needs a normalization rule** — trailing slashes,
+  percent-encoding, drive-letter case, symlinks. Without one, two hosts compute
+  different keys for the same server and neither is wrong (§10).
+- **Don't garbage-collect bindings whose consumer is absent.** A workspace that
+  isn't open contributes no customizations, which is not the same as the server
+  being deleted; collecting on absence would wipe bindings for every project a
+  user hasn't opened lately.
 
-1. **Renames and moves lose it.** Renaming the server, moving the manifest,
-   cloning to a different path, opening through a devcontainer, or bumping a
-   versioned plugin URL all change the key. It fails **safe** — no match, host
-   default — but invisibly, and a user who set a preference months ago won't
-   connect the two events.
-
-2. **URI equality is unspecified.** Trailing slashes, percent-encoding, Windows
-   drive-letter case, UNC paths, symlinks. Two implementations can compute
-   different keys for one server and neither is wrong. This needs a normalization
-   rule in the spec.
-
-3. **`name` uniqueness is a rule this proposal adds**, not one it inherits. Worth
-   being honest that the key's soundness rests on it.
-
-4. **Name reuse can still mis-bind.** Delete a `github` server bound to work,
-   later add a *different* `github` server at the same manifest pointing at
-   personal, and the old binding matches — silently using the **work** account.
-   Same class as the counter case above: much rarer, and user-caused rather than
-   host-caused, but not gone.
-
-   The tempting fix — delete bindings whose consumer isn't present — is **wrong**.
-   A workspace that isn't open contributes no customizations, and that isn't the
-   same as the server being deleted; collecting on absence would wipe bindings for
-   every project a user hasn't opened lately. The honest fix is **visibility**:
-   clients SHOULD show the binding ("github → tyler@work"), so a wrong one is
-   something a user can see and correct.
-
-#### Why accept it
-
-Problem 1 fails safe, 2 and 3 close by writing rules down, and 4 is rare,
-user-triggered, and visible in UI. Against that, the pair needs no change to a
-shared type used by every client, so it can't stall. It's also
-**forward-compatible**: if `CustomizationBase` later gains a real `stableId`
-(§10), bindings migrate by matching `(uri, name)` once and rewriting to the new
-key — no flag day, no lost preferences.
+The residual failure is a rename or move losing the binding. That fails **safe**
+— no match, host default — and clients SHOULD render the binding
+("*server* → *account*") so a wrong or missing one is visible rather than
+silent. If `CustomizationBase` later gains a real `stableId` (§10), bindings
+migrate by matching `(uri, name)` once and rewriting to the new key.
 
 ---
 
@@ -580,13 +550,14 @@ behaves exactly as today.
 
 | Symbol | File | Kind |
 | --- | --- | --- |
-| `InitializeResult.hostAuth?` | `common/commands.ts` | added (capability) |
-| `HostAuthCapability` | `common/commands.ts` | added (type) |
+| `InitializeResult.authentication?` | `common/commands.ts` | added (capability) |
+| `AuthenticationCapability` + `AuthFlowSupport` | `common/commands.ts` | added (types) |
 | `AuthFlowKind` | `common/state.ts` | added (enum) |
+| `AuthFlowRequest` | `channels-accounts/commands.ts` | added (union; carries `redirectUri`) |
 | `authBegin` / `authComplete` | `channels-accounts/commands.ts` | added (2 commands) |
-| `AccountsState`, `HostAccount` (identity), `AuthChallenge` (need), `AuthAttempt` (try) | `channels-accounts/state.ts` | added (state) |
-| `AccountConsumer` (agent variant carries `resource?`) + `HostAccount.consumers?` | `channels-accounts/state.ts` | added (multi-account) |
-| `AccountOrigin`, `AuthAttemptStatus` | `channels-accounts/state.ts` | added (enums) |
+| `AccountsState`, `HostAccount` (identity), `AuthChallenge` (need), `AuthAttemptState` (try) | `channels-accounts/state.ts` | added (state) |
+| `AccountConsumer` (agent variant carries `resource`) + `HostAccount.consumers?` | `channels-accounts/state.ts` | added (multi-account) |
+| `AuthAttemptStatus` + 5 `AuthAttempt*State` variants | `channels-accounts/state.ts` | added (lifecycle union) |
 | `DeviceCodePrompt` | `channels-accounts/state.ts` | added (device-code payload) |
 | `McpAuthRequiredReason` | `channels-session/state.ts` | reused (see §10 on the name) |
 | `accounts/set` + `accounts/updated` + `accounts/removed` | `channels-accounts/actions.ts` | added (keyed collection) |
@@ -598,7 +569,7 @@ behaves exactly as today.
 | 7 `ActionType` entries + `ACTION_INTRODUCED_IN` | `common/actions.ts`, `version/registry.ts` | added |
 
 > **On shape.** All three collections here (`accounts`, `challenges`,
-> `pendingAuth`) are keyed, so each uses the `Set` / `Updated` / `Removed`
+> `attempts`) are keyed, so each uses the `Set` / `Updated` / `Removed`
 > convention rather than a single full-replacement action — the
 > `root/terminalsChanged` full-replacement style is a catalogue exception, not
 > the convention. Each action needs a fixture per reducer branch (insert,
@@ -613,11 +584,20 @@ behaves exactly as today.
 interface InitializeResult {
   // …existing…
   /** Present = this host owns credentials and accepts the auth commands. */
-  hostAuth?: HostAuthCapability;
+  authentication?: AuthenticationCapability;
 }
-interface HostAuthCapability {
+interface AuthenticationCapability {
   /** The menu. The client picks from it in `AuthBeginParams.flows`. */
-  flows: AuthFlowKind[];
+  flows: AuthFlowSupport[];
+}
+
+/**
+ * An object rather than a bare `AuthFlowKind`, so a flow can later advertise
+ * facts about itself — supported PKCE methods, a required redirect scheme, a
+ * polling floor — without reshaping the capability.
+ */
+interface AuthFlowSupport {
+  kind: AuthFlowKind;
 }
 
 // ── Flows — common/state.ts ──────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


TylerLeonhardt:
[ahp-host-owned-auth.html](https://github.com/user-attachments/files/31201488/ahp-host-owned-auth.html)

roblourens (dismissed):

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+    C->>U: opens the browser / displays the code
+    U->>AS: authenticates and consents
+    Note over C,H: authorization-code flow only
+    C->>H: authComplete({ attemptId, callbackUri })
+    H->>AS: token exchange (host-side)
+    H->>C: accounts/set + accounts/challengeRemoved
+```
+
+**Where `challengeId` comes from.** Step 2, and nowhere else. The host makes
+`AuthChallenge.id` and posts it with `accounts/challengeSet`. Any client
```
connor4312, line 219: Do we still have the AuthRequired error? Does a client need to have subscribed to that channel to know that auth is pending for the account?

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+
+Bindings live in accounts state rather than at each consumption site, because
+that is what an account-management UI renders; deriving it by walking every
+session's customizations is both expensive and racy. Per-session overrides and
+`usedBy` rollups stay deferred (§10).
+
+### 7.2 The consumer key: a real gap in AHP today
+
+A binding has to survive a restart, so it needs an id for "this MCP server" that
+means the same thing in every session. **AHP doesn't guarantee one.**
```
connor4312: Eh, this is more just up to the host to implement correctly. Hosts will want to have auth grants maintained for an auth server over restarts to have a good experience, but that is already something they have to do internally anyway.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+
+### 8.2 Type signatures
+
+```ts
+// ── Negotiation — common/commands.ts ─────────────────────────────────────
+// One-sided: the host advertises, the client picks. No client capability (§9).
+interface InitializeResult {
+  // …existing…
+  /** Present = this host owns credentials and accepts the auth commands. */
+  hostAuth?: HostAuthCapability;
```
connor4312:
```suggestion
  authentication?: AuthenticationCapability;
```

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+}
+interface HostAuthCapability {
+  /** The menu. The client picks from it in `AuthBeginParams.flows`. */
+  flows: AuthFlowKind[];
+}
+
+// ── Flows — common/state.ts ──────────────────────────────────────────────
+const enum AuthFlowKind {
```
connor4312, line 604: Do you forsee the need for additional flow-specific details here? If so it might make sense to have flows be `flows: { kind: AuthFlowKind }[]` for forward compatibility

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+interface HostAccount {
+  /** Stable, host-assigned. */
+  id: string;
+  /** The authorization server that authenticated the user. Identity comes from
+   *  the issuer, not from any resource — hence its place here. */
+  issuer?: string;
+  /** Stable subject at that issuer, when known. */
+  subject?: string;
+  /** Display only, e.g. "tyler@github.com". What the user picks between. */
+  label?: string;
```
connor4312: I think we'd probably want to require the label, otherwise clients have no way to display this to users, right?

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+  /** How the host came to hold it. Decides revocation behaviour (§8.4). */
+  origin?: AccountOrigin;
```
connor4312: Do we actually want to express 'origin' or do we just want to express what operations a client can do on the account?

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+ * A consumer value identifies *what uses a credential*, which for a
+ * multi-resource agent is not the agent alone — see `resource` below.
+ */
+type AccountConsumer =
+  | {
+      kind: 'agent';
+      provider: string;
+      /**
+       * Which resource this binding is for. **Required**, not optional: one
+       * Copilot can hold a GitHub account and an Entra account at once, so
```
connor4312: We should not talk about specific providers like this in the spec

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+       * credential already names one, and `ProtectedResourceMetadata.resource`
+       * is itself REQUIRED:
+       *
+       * - declared — `AgentInfo.protectedResources[].resource` (root channel)
+       * - dynamic — `AuthRequiredErrorData.resources[].resource` (`-32007`)
+       * - pushed — `AuthenticateParams.resource`
+       *
+       * So the client still only echoes host-published data.
+       */
+      resource: string;
```
connor4312, line 748: Is this basically an opaque ID for clients, or does this mean something? E.g. for copilot this would I assume be whatever the copilot endpoint is, but is that meaningful?

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+   *  though, not one forever: a server can later declare a different
+   *  identifier, which is why a stored copy is a staleness hazard rather than
+   *  merely redundant.
+   *
+   *  AHP already draws this line. Agent auth needs are plural everywhere they
+   *  appear — `AgentInfo.protectedResources` and `AuthRequiredErrorData
+   *  .resources` are both arrays — while `McpAuthRequirement.resource`, shared
+   *  by `McpServerAuthRequiredState` and `ToolCallAuthRequiredState.auth`, is a
+   *  single object. These variants mirror that, rather than inventing it. */
+  | { kind: 'mcpServer'; uri: URI; name: string };
```
connor4312: Similarly here, does the MCP server URI mean anything to clients or is the name sufficient?

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+
+/**
+ * One in-flight sign-in. Discriminated on `flow`, so device-code facts exist
+ * only where they mean something — `{ flow: AuthorizationCode, device: {…} }`
+ * can't be said, and neither can a device-code attempt with nothing to show.
+ */
+type AuthAttempt = AuthorizationCodeAttempt | DeviceCodeAttempt;
+
+interface AuthAttemptBase {
+  attemptId: string;
```
connor4312:
```suggestion
  id: string;
```

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+  resource: string;
```
connor4312, line 811:
```suggestion
  resource: URI;
```

I assume?

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+  /** The need this try is answering. Absent for flow 4 (§5.3). */
+  challengeId?: string;
+  status: AuthAttemptStatus;
+  /** Set once `status` is `Completed`. */
+  accountId?: string;
+  /** Set once `status` is `Failed`. */
+  error?: ErrorInfo;
```
connor4312: THis should be expressed via a union type descriminated by AuthAttemptStatus

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+  // Nothing to broadcast: the browser is already open on the client that
+  // called `authBegin`, and the URI is bound to that client's `redirectUri`.
+}
+
+interface DeviceCodeAttempt extends AuthAttemptBase {
+  flow: AuthFlowKind.DeviceCode;
+  /** Required — showing this is the whole reason `pendingAuth` exists. */
+  device: DeviceCodePrompt;
+  /** RFC 8628 `expires_in` as an absolute instant. Required: the code stops
+   *  working, and a client showing it has to be able to say when. */
```
connor4312:
```suggestion
  /** RFC 8628 `expires_in` as an ISO 8601 timestamp. Required: the code stops
   *  working, and a client showing it has to be able to say when. */
```

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+}
+
+interface ChallengeSetAction {        // keyed by AuthChallenge.id
+  type: ActionType.ChallengeSet;      // 'accounts/challengeSet'
+  challenge: AuthChallenge;           // host-authored
+}
+interface ChallengeRemovedAction {
+  type: ActionType.ChallengeRemoved;  // 'accounts/challengeRemoved' — satisfied
+  challengeId: string;
+}
```
connor4312, line 916: for consistency, double check in our actions if we normally prefix it like `accountId` or just have it be `id` and "account" is contextual from the action

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1409 @@
...
+   *
+   * Echoing is a guarantee, not a convention: the host MUST reject a value it
+   * did not publish with `InvalidParams` (`-32602`) — see §8.4.
+   */
+  target: { challengeId: string } | { consumer: AccountConsumer };
+  /** Flows the client will drive for THIS attempt, most-preferred first. */
+  flows: AuthFlowKind[];
+  /** REQUIRED when `flows` includes `authorizationCode`. Loopback per
+   *  RFC 8252 §7.3 unless host policy says otherwise. */
+  redirectUri?: URI;
```
connor4312: Can be a union such that redirectUri is required along with the authorization code flow request. Makes invalid states harder to express. General instruction for agent, if it thinks a property is "Required when" or "Available when", it should consider expression that as a more descriptive union type.

roblourens (approved):

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+remote hosts and absent clients. Authorization-code-with-loopback needs the client
+to run an HTTP server *and* sit next to the browser. Device code needs "show a
+string and a URL" — every client can do that, including as an
+[`InputRequestResponsePart`](/guide/elicitation). No `redirectUri`, no listener,
+no callback relay, and the polling stays host-side so the result just lands in
+state. Device code is the baseline; authorization code is the optimization for
+local rich clients.
+
+**PKCE stays host-side, as a MUST.** The authorization code passes through the
+client. That's only acceptable because the verifier doesn't, which makes the code
```
dmitrivMS, line 407: AI Review: The client receives the authorization URI and controls the browser/loopback callback, so it can replace the host's `code_challenge` with one derived from its own verifier while preserving `state`, `client_id`, and `redirect_uri`. PKCE binds the code to the challenge received by the authorization server, so a public relay client can then redeem the code itself; validating `state` does not detect this. Require authorization-server-enforced request integrity such as signed JAR or PAR with direct requests disabled, require token-endpoint authentication unavailable to the client, or omit this relay flow.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+- **URI comparison needs a normalization rule** — trailing slashes,
+  percent-encoding, drive-letter case, symlinks. Without one, two hosts compute
+  different keys for the same server and neither is wrong (§10).
+- **Don't garbage-collect bindings whose consumer is absent.** A workspace that
+  isn't open contributes no customizations, which is not the same as the server
+  being deleted; collecting on absence would wipe bindings for every project a
+  user hasn't opened lately.
+
+The residual failure is a rename or move losing the binding. That fails **safe**
+— no match, host default — and clients SHOULD render the binding
```
dmitrivMS, line 536: AI Review: This is not a safe failure. After a bound MCP declaration is renamed or moved, `(uri, name)` no longer matches and the host-default fallback can select a different identity. Since this feature supports work/personal accounts and unattended runs, the server can continue under the wrong account without a client present. Keep an unresolved explicit binding in an auth/input-needed state until it is rebound, or migrate it through a stable identifier.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+   * `false` for a credential the host did not mint and cannot forget — an
+   * ambient one it rediscovers from its environment on every start, or one a
+   * client pushed and still owns. Signing those out is not the host's to do,
+   * so a client SHOULD NOT offer the affordance.
+   *
+   * This is deliberately *what a client can do*, not *where the credential
+   * came from*. Provenance is host-internal and drives the host's own
+   * revocation rules (§8.4); the only part of it a client can act on is this.
+   */
+  removable: boolean;
```
dmitrivMS, line 701: AI Review: `HostAccount` combines one identity's credentials across resources, but removal behavior differs by credential origin: host-owned grants are removable, while ambient and pushed credentials are not. If one identity contains both origins, neither boolean value is correct. Model removal at grant/origin granularity, or prohibit mixed-origin aggregation and define how those credentials are represented as separate accounts.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+/**
+ * A union, not a bare string, so "an id of unspecified kind" can't be said.
+ *
+ * A consumer value identifies *what uses a credential*, which for a
+ * multi-resource agent is not the agent alone — see `resource` below.
+ */
+type AccountConsumer =
+  | {
+      /** Matches `AgentInfo.provider`. */
+      provider: string;
```
dmitrivMS, line 723: AI Review: The agent arm has no `kind`, while the MCP arm uses `kind: 'mcpServer'` and the proposal later defines this wire type as a discriminated union with `kind: 'agent' | 'mcpServer'`. Generated clients need a discriminator shared by every variant to decode the union consistently. Add `kind: 'agent'` to this arm.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+interface AccountSetAction {          // keyed by HostAccount.id
+  type: ActionType.AccountSet;        // 'accounts/set'
+  account: HostAccount;               // host-authored
+}
+interface AccountUpdatedAction {
+  type: ActionType.AccountUpdated;    // 'accounts/updated'
+  id: string;
+  /** The only field a client may change, so binding is one action — the host
+   *  reconciles by dropping the consumer from whoever held it before. */
+  consumers?: AccountConsumer[];
```
dmitrivMS, line 899: AI Review: Replacing one account's complete `consumers` array loses concurrent edits from clients that derived updates from the same snapshot. A cross-account move also cannot be atomic: flow 4 publishes `accounts/set` before removing the consumer from the old account, exposing duplicate ownership. Define a consumer-keyed move/delta action whose reducer removes the prior binding and adds the destination in one transition.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+   * Flows the client will drive for THIS attempt, most-preferred first.
+   *
+   * Each entry carries what its own flow needs, so there is no field that is
+   * "required when some other field says so" — offering the authorization-code
+   * flow without a `redirectUri` is not expressible, and supplying one without
+   * offering that flow is not either.
+   */
+  flows: AuthFlowRequest[];
+  /**
+   * Act on an EXISTING account — re-authorizing after expiry, or consenting to
```
dmitrivMS, line 961: AI Review: `accountId` says this flow renews or widens an existing identity, but an account chooser can return credentials for another principal. Without a post-exchange identity check, the host can store that grant under the old ID and silently switch every existing consumer binding. Require the resulting stable issuer/subject or provider identity to match; otherwise fail or create a distinct unbound account.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+   * more. Absent means "add a new one" (flow 4). An optional id rather than an
+   * `intent` enum, so "an existing account, unspecified which" can't be said.
+   * Defaults to a challenge's own `accountId` when it has one.
```
dmitrivMS, line 964: AI Review: Omitting `accountId` has contradictory meanings: the preceding text says it adds a new account, while this line defaults the omission to the challenge's existing account. An expired or insufficient-scope challenge therefore has no explicit way to choose a different identity. Use a discriminated new-versus-existing intent so both operations remain expressible.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+}
+```
+
+### 8.3 Versioning & gating
+
+- Purely additive, so it rides the next **MINOR** (`0.9.0`, or `0.8.0` if it
+  lands before that release is cut). No breaking window needed.
+- The seven actions each carry an `ACTION_INTRODUCED_IN` entry. Five are
+  host-authored (`accounts/set`, both `challenge*` actions, both
+  `authAttempt*` actions); `accounts/updated` and `accounts/removed` are
```
dmitrivMS, line 1034: AI Review: This classifies both `authAttempt*` actions as host-authored, but the proposal specifies that clients cancel sign-in by dispatching `accounts/authAttemptRemoved`. Dispatch validation rejects actions without `@clientDispatchable`, so the documented cancellation cannot be sent. Mark removal client-dispatchable with validation semantics, or expose a cancellation command.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+  gated by `InitializeResult.authentication` plus the `initialize` version handshake —
+  a one-sided advertisement, with no matching client capability.
+- A host MUST NOT advertise `InitializeResult.authentication` to a connection it is
+  unwilling to let spend its credentials, and MAY reject
+  `subscribe('ahp-accounts://')` with `PermissionDenied` (`-32009`) even when it
+  does.
+- Every remove action is idempotent and a no-op against an already-settled
+  target, per the keyed-collection convention. Because client-dispatched actions
+  are notifications with no response, a rejected mutation surfaces as the host
+  simply not echoing it — the standard write-ahead reconciliation model rather
```
dmitrivMS, line 1045: AI Review: Silently withholding a rejected action contradicts AHP write-ahead reconciliation. A client removes its optimistic mutation from `pendingActions` only after receiving the matching ordered envelope; rejection should echo the action and origin with `ActionEnvelope.rejectionReason` without applying it to confirmed state. Otherwise the rejected mutation remains pending and is repeatedly rebased.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+### 8.6 How the host works out scopes
+
+`authBegin` carries no `scopes`, so the host works them out from what it already
+holds. The chain isn't new — it's the one `McpAuthRequirement.requiredScopes`
+already documents, *"parsed from the `WWW-Authenticate` header (or
+`scopes_supported` fallback)"*:
+
+| Target | Host uses |
+| --- | --- |
+| `{ challengeId }` (flows 1–3) | that challenge's `requiredScopes`, falling back to its `resource.scopes_supported` |
```
dmitrivMS, line 1161: AI Review: This applies to proactive first-run challenges where optional `requiredScopes` may be absent, so the fallback requests every value in `scopes_supported`. RFC 9728 describes that field as a capability catalogue, not an instruction to request all scopes. Require the exact least-privilege scopes for static challenges, or use the authorization server's documented default; do not request the whole catalogue.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+
+- **Attempts in state, authorization URI not** (§7). Device codes are
+  broadcastable and belong in state; the authorization URI is tied to one client's
+  redirect URI and stays in the `authBegin` result.
+
+- **Its own channel, not a root-state field** (§7). Seeing accounts should be
+  refusable separately from seeing root state.
+
+- **Pushed tokens MAY appear as accounts.** A host that also accepts
+  `authenticate` MAY surface them so a client sees one picture. They are
```
dmitrivMS, line 1410: AI Review: `authenticate` credentials are per-connection, while `ahp-accounts://` is singleton state shared across clients. Publishing a pushed credential here exposes an account and permits global bindings without recording which connection owns it, whether peers can use it, which credential wins when a host-owned binding also exists, or what happens on disconnect. Keep pushed credentials out of this channel or model connection-scoped ownership, visibility, lifetime, precedence, and binding rules.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+  Everything that decides it is host-side: whether the host can store a refresh
+  credential safely, whether it's sandboxed (in which case it should refuse one
+  regardless of what a client asks), and whether it runs unattended work at all.
+  The user's control is already in the right place — asking for offline access
+  changes the consent screen.
+
+- **Challenges are state on the accounts channel** (§5.2). Rejected scraping
+  `challengeId` out of `McpServerCustomization.state` and `toolCall.auth`, where
+  it's missing entirely for agent-level `-32007`. A host-authored list makes the
+  channel self-sufficient, and entries are **deduplicated** — two sessions blocked
```
dmitrivMS, line 1425: AI Review: Two sessions using the same MCP server can be blocked by different simultaneous `requiredScopes`. Coalescing them into one server-level challenge must either discard one requirement or union both, while section 8.6 forbids unioning unrelated pending demands. Deduplicate over the exact authorization need, including resource, target account, and scope set, so disjoint requirements remain independently actionable.

docs/proposals/host-owned-auth.md
```diff
@@ -0,0 +1,1532 @@
...
+  enforces it while sequencing; whether that deserves stronger protocol treatment
+  is open.
+
+- **Per-session account selection.** A standing binding (§7.1) covers the common
+  case. The minimal answer is an optional per-resource account map on
+  `CreateSessionParams`; the fuller one interacts with per-chat config
+  ([#335](https://github.com/microsoft/agent-host-protocol/issues/335)).
+
+- **Surfacing host-default bindings.** `consumers` covers what a user explicitly
+  bound, so a client can warn "signing out breaks these 3 servers." It doesn't
```
dmitrivMS, line 1490: AI Review: Flow 4 requires the client to copy a `consumer` from `AccountsState`, and the host must reject targets it did not publish. Because default-bound consumers are omitted from `HostAccount.consumers`, a working consumer using the host default has no legal `authBegin` target and cannot perform the documented switch-account flow. Publish effective default bindings or another host-authored consumer catalogue.

connor4312 (approved):

PR #411, opened 2026-08-23
Harden .NET client security, reconnects, and Native AOT

davidfowl:
## Summary

- create Unix client-ID temporary files with owner-only permissions before writing any bytes
- preserve atomic replacement and Windows behavior across `net8.0` and `netstandard2.0`
- tighten caller-owned configuration, serializer, value-object, and invalid-state API boundaries before first publication
- validate negotiated protocol versions and snapshot mutable WebSocket options before connection callbacks
- make multi-host reconnects preserve live subscriptions and the highest applied server sequence, publish replay/snapshot state before `Connected`, and serialize connection replacement with subscription bookkeeping
- prevent cancelled or non-cooperative connection attempts from leaking or overwriting the committed client
- generate System.Text.Json metadata for the complete protocol model and remove reflection from union and reducer dispatch
- publish and execute a reflection-disabled Native AOT smoke application that calls the real `PingAsync` API
- add `ClientConfig.TimeProvider` so request timeouts, keep-alive, reconnect scheduling, and host timestamps are deterministic in tests
- prevent timed-out queued requests from being transmitted without canceling active WebSocket writes and aborting the shared connection

## Security

`FileClientIdStore` no longer creates a default-permission temporary file and chmods it after content is written. On Unix:

- `net8.0` creates the file with `FileStreamOptions.UnixCreateMode = 0600`
- `netstandard2.0` uses native `open(..., O_CREAT | O_EXCL | O_CLOEXEC, 0600)` followed by `fchmod(0600)` before exposing the stream
- unknown Unix platforms fail closed rather than guessing an `O_CLOEXEC` value
- failures occur before client-ID bytes are written and temporary files are cleaned up

Regression coverage exercises both paths on Linux and verifies the temporary file is empty and exactly owner-readable/writable at the pre-write boundary.

## API and runtime review

The complete public surfaces of `Microsoft.AgentHostProtocol` and `Microsoft.AgentHostProtocol.Abstractions` were inspected for both target frameworks and checked against examples and tests. The two TFMs expose matching APIs. High-confidence corrections include configuration ownership, frozen serializer options, valid `StringOrMarkdown` construction, `HostId` value semantics, accidental visibility, protocol negotiation validation, connection cleanup, duplicate host registration, reconnect correctness, and deterministic time control.

The runtime review also exposed reconnect gaps now covered by regression tests: dynamic subscription replay, live sequence advancement only after successful application, snapshot reconnect without re-initialization, snapshot delivery through `SubscriptionEventSnapshot`, replay `missing` pruning while retaining stateless subscriptions, handshake event buffering, root-action mirroring, old-pump draining, replacement readiness, post-ack subscription races, commit/abandon arbitration, and cleanup when a transport factory ignores cancellation.

## Native AOT

The default serializer now uses generated metadata for the full canonical wire model. JSON-RPC shape converters, discriminated unions, snapshots, tool inputs, reducer action metadata, and the closed dictionary used by `PingAsync` all use generated metadata rather than runtime serialization/reflection. The generated context remains internal so its generated members do not become compatibility surface; consumers can compose the single `AhpJsonMetadata.Default` resolver API instead.

Normal reflection-enabled .NET applications retain compatibility for application-defined values passed through generic serializer/client APIs. Reflection-disabled or Native AOT applications provide their own source-generated resolver for custom values. CI publishes and runs a reflection-disabled native executable that exercises initialization, framing, unions, wire enums, snapshots, scalar unions, and a real client ping round trip.

## Intentionally deferred

- Generated union wrappers still expose parameterless construction and mutable `AhpUnion.Value`; correcting this safely requires coordinated generator/converter changes across the generated model surface.
- Server-request handler delegates do not accept cancellation tokens; adding them is a cross-cutting public API design change.
- Unknown/malformed notifications are skipped rather than surfaced through a raw-notification or diagnostic API.
- The canonical generated package intentionally exposes the complete wire model; splitting that model would trade compatibility surface for package/versioning complexity and needs a separate design decision.
- Generated literal-discriminant defaults and the legacy session-action/`ActionType` gap are tracked separately in #414 rather than duplicated here.

## Validation

- `dotnet format whitespace --verify-no-changes --no-restore`
- Release builds for `netstandard2.0` and `net8.0`
- 524 .NET tests on Windows
- reflection-disabled Native AOT publish and execution on Windows
- NuGet packing for both packages and both TFMs
- `npm test` (typecheck, lint, release/changelog/generated checks, TypeScript tests, and 100% reducer coverage)
- generated-source and changelog-fragment freshness checks
- iterative final code review with no remaining findings


connor4312 (dismissed):

joshmouch:
One thing I noticed reading through the generator changes here — there's a discriminant-default bug in `generate-csharp.ts` that predates this PR and that this PR doesn't touch. Fix is up as #414; flagging it here since you're in that file.

`literalValue` is computed at `:338`/`:343` and stored at `:368`, but it's never read — `grep -n literalValue scripts/generate-csharp.ts` returns only those declaration and write sites. So the generated types emit their discriminant with no initializer:

```csharp
public ActionType Type { get; init; }
```

and since `RootAgentsChanged` is the first member of `ActionType`, `default(ActionType)` is `root/agentsChanged`:

```csharp
var action = new StateAction(new ChatDraftChangedAction { /* ... */ });
// serializes as {"type":"root/agentsChanged", ...}
```

`UnionConverter.Write` serializes by the runtime type precisely so that "every property (including the variant's own discriminator field) is written", so the record's own `Type` is what reaches the wire. 94 records across `Actions`, `State` and `Commands`.

What makes it .NET-specific isn't the missing initializer, it's that C#'s zero value is a *valid* member. Rust can't express it — `StateAction` is `#[serde(tag = "type")]` over an enum, so the discriminant rides on the variant. Swift and Kotlin are the same shape. Go can, but `type ActionType string` zero-values to `""`, which isn't a valid discriminant, so it fails loudly. C# is the only one where the default is well-formed and wrong, so it's the only one that mis-tags silently.

#414 keeps clear of your serializer/AOT work — I checked it applies on top of this PR's head, where the defect is also present.

One thing I left out of #414 because it's your call rather than a patch: the 16 hand-written session actions in `SESSION_ACTION_TYPES_CS` have wire values that aren't members of `ActionType` at all (`session/turnStarted`, `session/delta`, `session/error`, …). They deserialize fine — the variant map keys on the wire string — but there's no enum member to serialize back out, so reading one and writing it returns `root/agentsChanged`. That's a round-trip break rather than a default bug, and fixing it means either adding those members to `ActionType` or moving those records to a plain `string` discriminant. Happy to implement whichever you'd prefer.

Worth noting neither is reachable by CI today: zero of the 42 fixtures in `types/test-cases/round-trips/` cover any of those 16 wire values.

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
```diff
@@ -795,19 +873,32 @@ private ulong ApplyReconnectReplay(HostEntry entry, ReconnectReplayResult replay
...
     /// </summary>
-    private static RootState? ExtractRootSnapshot(InitializeResult result)
+    private static ulong ApplyReconnectSnapshot(HostEntry entry, ReconnectSnapshotResult result)
+    {
+        var lastSeq = entry.Snapshot().ServerSeq;
+        foreach (var snapshot in result.Snapshots)
+        {
+            if (snapshot.FromSeq > lastSeq) lastSeq = snapshot.FromSeq;
+        }
+        return entry.ApplyConnected(ExtractRootSnapshot(result.Snapshots), lastSeq);
```
dmitrivMS: AI Review: `ApplyReconnectSnapshot` advances the sequence and applies only the root snapshot; every returned session, chat, terminal, or changeset snapshot is discarded before the host is marked connected. Because the .NET surface exposes neither the reconnect result nor a snapshot event, consumers cannot reset their reducers after a replay gap and will continue from stale state. Publish or apply every returned snapshot before reporting the host connected.

scripts/generate-csharp.ts
```diff
@@ -2648,6 +2657,89 @@ function generateTelemetryFile(project: Project): string {
   return `${fileHeader()}\n${body.join('\n')}\n`;
 }
 
+function generateJsonSerializerContext(generatedFiles: readonly string[]): string {
+  const serializableTypes = new Set<string>(['StringOrMarkdown']);
```
dmitrivMS: AI Review: `PingAsync` serializes a library-owned `Dictionary<string, string>`, but this generated context roots only `StringOrMarkdown` plus generated declarations. In a reflection-disabled Native AOT app, the default serializer therefore throws `NotSupportedException` before sending the ping. Explicitly root this closed dictionary type or introduce a generated `PingParams`, and exercise `PingAsync` in the AOT smoke test.

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
```diff
@@ -1440,21 +1587,31 @@ public async Task<SubscribeResult> SubscribeAsync(
...
                 ?? throw new AhpRpcException(JsonRpcErrorCodes.InternalError, "ahp: subscribe returned no result");
-            entry.AppendSubscription(uri);
-            return result;
-        }
-        catch
-        {
-            sub.Dispose();
-            throw;
+
+            await entry.ConnectionGate.WaitAsync(cancellationToken).ConfigureAwait(false);
```
dmitrivMS: AI Review: After `subscribe` is acknowledged, this second gate wait still uses the caller token. If reconnect owns the gate when that token is canceled, the method throws before `AppendSubscription`, leaving the peer subscribed while the reconnect replay set omits the URI. `UnsubscribeAsync` has the symmetric problem and can later restore an already-removed subscription. Once the remote operation succeeds, commit replay bookkeeping independently of late caller cancellation; honor cancellation only before retrying on a replacement connection.

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
```diff
@@ -632,10 +676,18 @@ public async Task ShutdownAsync(CancellationToken cancellationToken = default)
             // cancellation and exits rather than blocking on its manual-reconnect
             // wait forever.
             entry.SignalManualReconnect();
-            var client = entry.CurrentClient;
-            if (client is not null)
+            await entry.ConnectionGate.WaitAsync(CancellationToken.None).ConfigureAwait(false);
```
dmitrivMS, line 741: AI Review: `OpenHostAsync` now holds `ConnectionGate` while awaiting the caller-supplied transport factory and handshake, but shutdown waits for that gate with `CancellationToken.None`. A factory that ignores cancellation can therefore block shutdown forever, even when the shutdown token is canceled. Avoid holding the gate across arbitrary connection setup, or let teardown abandon and safely discard a stale attempt.

clients/dotnet/src/AgentHostProtocol/AhpClient.cs
```diff
@@ -496,27 +529,23 @@ await ShutdownWithErrorAsync(
...
+                    item.Sent?.TrySetCanceled(item.CancellationToken);
+                    continue;
+                }
+
                 var frame = _serializer.EncodeMessage(item.Message);
                 try
                 {
+                    // Canceling ClientWebSocket.SendAsync aborts the shared connection,
+                    // so message cancellation only prevents writes that have not started.
                     await _transport.SendAsync(frame, _lifetimeCts.Token).ConfigureAwait(false);
```
dmitrivMS, line 556: AI Review: The request timeout token is checked only before synchronous `EncodeMessage`. If an injected serializer blocks past the timeout, the caller observes cancellation and the pending request is removed, but the writer then sends the frame using only the client lifetime token. The server can execute a request after the caller was told it timed out. Recheck the item token after encoding and skip the transport send when it has fired.

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
```diff
@@ -670,83 +722,108 @@ public async ValueTask DisposeAsync()
...
+                        // to a fresh `initialize` on the still-live client below. A
+                        // cancellation (shutdown/dispose) is NOT swallowed: it propagates
+                        // so the supervisor tears down promptly instead of blocking on a
+                        // fallback initialize.
+                    }
+
+                    if (reconnectResult?.Value is ReconnectReplayResult replay)
+                    {
+                        await SeedSessionSummariesAsync(entry, client, cancellationToken).ConfigureAwait(false);
+                        _ = ApplyReconnectReplay(entry, replay);
```
dmitrivMS: AI Review: `ApplyReconnectReplay` writes replayed actions to public channels before `SetClient` installs the replacement. The supervisor has already cleared `CurrentClient`, so a consumer reacting to an early replay event can call `DispatchAsync` and receive `HostNotConnectedException` during a successful reconnect. Install the replacement client before replay fan-out, or separate internal replay application from publication.

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
```diff
@@ -782,6 +858,8 @@ private ulong ApplyReconnectReplay(HostEntry entry, ReconnectReplayResult replay
             {
                 var evt = new SubscriptionEventAction(env);
                 ApplyEventToHostState(entry, evt);
+                if (env.Channel == ProtocolVersion.RootResourceUri)
+                    entry.ApplyRootAction(env.Action);
```
dmitrivMS, line 1042: AI Review: `ApplyConnected` commits the maximum replay sequence before this newly added root-action application runs. A wire payload with a present but null `agents` value is accepted by the generated serializer, then throws here; the supervisor retries using the already-advanced cursor, allowing the server to omit the failed and remaining replay actions. Apply or validate envelopes before advancing `ServerSeq`, updating the cursor only through the last successfully applied action.

davidfowl: Fixed in d1d6ec68. Replay sequences are prevalidated for strict ordering, then each action is applied before its `ServerSeq` is committed. Live pumping follows the same ordering. The malformed-replay regression confirms a later reconnect resumes from the last successfully applied action rather than the replay maximum.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/AhpClient.cs
davidfowl: Fixed in d1d6ec68. The writer rechecks cancellation immediately after synchronous serialization and before queueing the payload. A deterministic `FakeTimeProvider` regression blocks serialization, advances through the timeout, then verifies the expired request is never sent.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/AhpClient.cs
```diff
@@ -52,8 +52,12 @@ private KeepAlivePolicy(bool isEnabled, TimeSpan interval, TimeSpan timeout)
...
     /// <c>.ping(interval:timeout:)</c>.
     /// </summary>
-    public static KeepAlivePolicy Ping(TimeSpan interval, TimeSpan timeout) =>
-        new(isEnabled: true, interval: interval, timeout: timeout);
+    public static KeepAlivePolicy Ping(TimeSpan interval, TimeSpan timeout)
+    {
+        if (interval <= TimeSpan.Zero)
+            throw new ArgumentOutOfRangeException(nameof(interval), interval, "Keep-alive interval must be positive.");
+        return new(isEnabled: true, interval: interval, timeout: timeout);
+    }
```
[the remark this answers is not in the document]
davidfowl: Fixed in d1d6ec68. `KeepAlivePolicy.Ping` now rejects zero and negative timeouts, and the focused policy test covers both invalid values.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. The replacement client is published before replay/snapshot delivery, so callbacks can dispatch immediately. Commit/abandon arbitration now also prevents a cancelled attempt from overlapping a later installation while synchronous replay publication finishes. The regression dispatches from the first event of a 500-action replay.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. Transport creation and handshake now run outside `ConnectionGate`. A cancelled non-cooperative attempt is observed in the background and retains ownership of eventual client/transport cleanup, while teardown can acquire the gate and complete promptly. The regression uses a factory that ignores cancellation and returns only after shutdown.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. Once the peer acknowledges subscribe/unsubscribe, bookkeeping uses the host lifetime rather than the caller token. If the connection changes after acknowledgement, the operation now waits for and repeats against the replacement; reconnect state and client readiness are updated atomically so the null-client handoff cannot fail spuriously.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. Reconnect snapshot results now publish every returned resource as `SubscriptionEventSnapshot` through both aggregate and per-resource streams before the host transitions to `Connected`. The README documents reducer replacement behavior and the regression covers a non-root session snapshot.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

scripts/generate-csharp.ts
davidfowl: Fixed in d1d6ec68. The generated serializer context now explicitly roots `Dictionary<string, string>`, and the reflection-disabled Native AOT smoke executable performs a real `AhpClient.PingAsync` request/response over a loopback transport.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. The replacement client is published before replay/snapshot delivery, so callbacks can dispatch immediately. Commit/abandon arbitration now also prevents a cancelled attempt from overlapping a later installation while synchronous replay publication finishes. The regression dispatches from the first event of a 500-action replay.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. Reconnect snapshot results now publish every returned resource as `SubscriptionEventSnapshot` through both aggregate and per-resource streams before the host transitions to `Connected`. The README documents reducer replacement behavior and the regression covers a non-root session snapshot.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

scripts/generate-csharp.ts
davidfowl: Fixed in d1d6ec68. The generated serializer context now explicitly roots `Dictionary<string, string>`, and the reflection-disabled Native AOT smoke executable performs a real `AhpClient.PingAsync` request/response over a loopback transport.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. Once the peer acknowledges subscribe/unsubscribe, bookkeeping uses the host lifetime rather than the caller token. If the connection changes after acknowledgement, the operation now waits for and repeats against the replacement; reconnect state and client readiness are updated atomically so the null-client handoff cannot fail spuriously.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. Replay sequences are prevalidated for strict ordering, then each action is applied before its `ServerSeq` is committed. Live pumping follows the same ordering. The malformed-replay regression confirms a later reconnect resumes from the last successfully applied action rather than the replay maximum.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/Hosts/MultiHostClient.cs
davidfowl: Fixed in d1d6ec68. Transport creation and handshake now run outside `ConnectionGate`. A cancelled non-cooperative attempt is observed in the background and retains ownership of eventual client/transport cleanup, while teardown can acquire the gate and complete promptly. The regression uses a factory that ignores cancellation and returns only after shutdown.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/AhpClient.cs
davidfowl: Fixed in d1d6ec68. The writer rechecks cancellation immediately after synchronous serialization and before queueing the payload. A deterministic `FakeTimeProvider` regression blocks serialization, advances through the timeout, then verifies the expired request is never sent.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

clients/dotnet/src/AgentHostProtocol/AhpClient.cs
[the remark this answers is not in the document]
davidfowl: Fixed in d1d6ec68. `KeepAlivePolicy.Ping` now rejects zero and negative timeouts, and the focused policy test covers both invalid values.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

dmitrivMS (approved):

connor4312 (approved):

PR #414, opened 2026-08-24
fix(dotnet): initialize generated literal discriminants

joshmouch:
## Summary

`scripts/generate-csharp.ts` computes each literal discriminant's value and then discards it. `literalValue` is assigned at `:338`/`:343` and stored on the descriptor at `:368`, and nothing ever reads it — `grep -n literalValue scripts/generate-csharp.ts` returns only those declaration and write sites.

So generated records emit their discriminant with no initializer:

```csharp
public ActionType Type { get; init; }
```

These are ordinary (non-`[Flags]`) enums whose zero value is a real member, so `default(ActionType)` is `ActionType.RootAgentsChanged`:

```csharp
var action = new StateAction(new ChatDraftChangedAction { /* ... */ });
// serializes as {"type":"root/agentsChanged", ...}
```

`UnionConverter.Write` serializes by the runtime type specifically so that *"every property (including the variant's own discriminator field) is written"*, so the record's own `Type` property is what reaches the wire. This is a silent mis-tag, not a cosmetic default.

**95 records** across `Actions.generated.cs`, `State.generated.cs` and `Commands.generated.cs` now carry their discriminant.

## Why this is .NET-specific

Not the missing initializer — the fact that C#'s zero value is a *valid* member.

- **Rust** can't express it: `StateAction` is `#[serde(tag = "type")]` over an enum, so the discriminant is carried by the variant and there is nothing to default. Swift and Kotlin have the same shape.
- **Go** can express it, but `type ActionType string` means the zero value is `""` — not a valid discriminant, so it fails loudly.
- **C#** is the only client whose default is a well-formed, wrong value.

## The fix

Carry the C# initializer expression alongside the wire value and emit it. The `required` modifier is dropped on an initialized discriminant — requiring callers to set a constant is the thing being fixed. Discriminants with no single correct value stay uninitialized by construction.

94 of the 95 are generated. The 95th, `ChatToolCallConfirmedAction`, is hand-written in the generator rather than generated — but unlike the session actions below it *does* have a matching `ActionType` member, so it is fixed by hand in the same block.

## Tests

`dotnet test`: **498 passed, 0 failed.**

`DiscriminantDefaultTests.EveryActionRecordDefaultsToItsOwnDiscriminant` reflects over every exported record with an `ActionType Type` property, derives the expected member from the record name, and asserts the default matches. Against the pre-fix generator it fails with 94 named records:

```
94 action record(s) serialize with the wrong discriminant:
  AnnotationsEntryRemovedAction.Type defaulted to RootAgentsChanged (wire 'root/agentsChanged')
    but should be AnnotationsEntryRemoved (wire 'annotations/entryRemoved')
  ...
```

`KnownMissingListDoesNotOutliveTheGapItDocuments` guards the exclusion list itself: it fails if `ActionType` ever gains members for the excluded names, so the list cannot silently keep a now-testable record out of coverage.

## A separate, deeper finding — needs a ruling, not a patch

**19 hand-written session action records** in `SESSION_ACTION_TYPES_CS` are deliberately not fixed here, because there is nothing correct to set them to. Their wire values are **not members of the `ActionType` enum at all**:

```
session/turnStarted   session/delta            session/responsePart
session/toolCallStart session/toolCallDelta    session/toolCallReady
session/toolCallComplete   session/toolCallResultConfirmed
session/toolCallConfirmed  session/toolCallContentChanged
session/turnComplete  session/turnCancelled    session/error
session/usage         session/reasoning        session/truncated
session/pendingMessageSet  session/pendingMessageRemoved
session/queuedMessagesReordered
```

They deserialize correctly, because the variant map keys on the wire string:

```csharp
["session/turnStarted"] = typeof(SessionTurnStartedAction),
```

But on the way out, `SessionTurnStartedAction.Type` is an `ActionType` with no member for `session/turnStarted`, so it serializes as `root/agentsChanged`. **Read a `session/turnStarted` action and write it back and its type changes** — a round-trip fidelity break, and a worse one than the defect this PR fixes.

The options are a protocol-level call rather than a client one: add the 19 members to `ActionType`, or move these records to a plain `string` discriminant, or something else you'd prefer. Happy to implement whichever you pick — deleting an entry from the exclusion list is what switches coverage on.

**Neither defect is reachable by the shared corpus.** Of the 42 fixtures in `types/test-cases/round-trips/`, **zero** cover any of those 19 wire values. Adding them is a `types/` change affecting every client, so it isn't in this PR — but it's probably the higher-leverage half, and I'd be glad to do it as a follow-up.

## Notes

- The generator plus its three regenerated outputs, one new test file, and a `### Fixed` changelog entry. `npm run generate:dotnet` is the only thing that touched the generated files.
- Orthogonal to the serializer/AOT work in #411 — the defect is present at that PR's head too, and this applies cleanly on top.

### DIFF
--- a/clients/dotnet/CHANGELOG.md
+++ b/clients/dotnet/CHANGELOG.md
@@ -11,18 +11,6 @@ this release negotiates.
 
 ## [Unreleased]
 
-### Fixed
-
-- Generated action, state and command records now initialize their literal
-  discriminant (`Type`/`Kind`/`Status`/`State`) to its own constant. Previously
-  the property was emitted with no initializer, so it fell back to
-  `default(ActionType)` — and because these are ordinary enums whose zero value
-  is a real member, an action constructed without explicitly setting `Type`
-  serialized with a valid but wrong discriminant (the first enum member,
-  `root/agentsChanged`). 94 records across `Actions`, `State` and `Commands`
-  were affected. The generator already computed the correct value and discarded
-  it.
-
 ### Added
 
 - **Multiroot working directories** (upstream
--- a/clients/dotnet/CHANGELOG.md
+++ b/clients/dotnet/CHANGELOG.md
@@ -11,6 +11,18 @@ this release negotiates.
 
 ## [Unreleased]
 
+### Fixed
+
+- Generated action, state and command records now initialize their literal
+  discriminant (`Type`/`Kind`/`Status`/`State`) to its own constant. Previously
+  the property was emitted with no initializer, so it fell back to
+  `default(ActionType)` — and because these are ordinary enums whose zero value
+  is a real member, an action constructed without explicitly setting `Type`
+  serialized with a valid but wrong discriminant (the first enum member,
+  `root/agentsChanged`). 94 records across `Actions`, `State` and `Commands`
+  were affected. The generator already computed the correct value and discarded
+  it.
+
 ### Added
 
 - **Multiroot working directories** (upstream
--- a/clients/dotnet/src/AgentHostProtocol.Abstractions/Generated/Actions.generated.cs
+++ b/clients/dotnet/src/AgentHostProtocol.Abstractions/Generated/Actions.generated.cs
@@ -1342,7 +1342,7 @@ public sealed record ChatToolCallReadyAction
 /// </summary>
 public sealed record ChatToolCallConfirmedAction
 {
-    public ActionType Type { get; init; }
+    public ActionType Type { get; init; } = ActionType.ChatToolCallConfirmed;
 
     public required string TurnId { get; init; }
 
--- a/clients/dotnet/src/AgentHostProtocol.Abstractions/Generated/Actions.generated.cs
+++ b/clients/dotnet/src/AgentHostProtocol.Abstractions/Generated/Actions.generated.cs
@@ -242,7 +242,7 @@ public sealed record ActionEnvelope
 /// <summary>Fired when available agent backends or their models change.</summary>
 public sealed record RootAgentsChangedAction
 {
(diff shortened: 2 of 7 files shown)


joshmouch:
Pushed a regression test — the original version of this PR had none, and since the suite was green both before and after the fix, that meant the defect simply wasn't covered.

`DiscriminantDefaultTests` reflects over every exported record with an `ActionType Type` property, derives the expected member from the record name, and asserts the default matches. Against the pre-fix generator it fails with **94 named records**:

```
94 action record(s) serialize with the wrong discriminant:
  AnnotationsEntryRemovedAction.Type defaulted to RootAgentsChanged (wire 'root/agentsChanged')
    but should be AnnotationsEntryRemoved (wire 'annotations/entryRemoved')
  ...
```

After the fix, green: 498 passed, 0 failed.

Writing it caught something I'd missed. **`ChatToolCallConfirmedAction` is hand-written** in the generator rather than generated — so my generator change didn't reach it — **but unlike the session actions it does have a matching `ActionType` member**, which means it was fixable. It's fixed now, so the count is 95 records carrying their discriminant rather than 94.

The 19 excluded records are the session actions whose wire values have no `ActionType` member at all (that's the separate finding in the PR description — one fewer than the 20 I first counted, because `ChatToolCallConfirmedAction` turned out to belong on the fixable side).

The exclusion list has its own guard: `KnownMissingListDoesNotOutliveTheGapItDocuments` fails if `ActionType` ever gains members for those names, so the list can't silently keep a now-testable record out of coverage. Whichever way you rule on the 19, deleting an entry is what turns coverage on.

PR #415, opened 2026-08-24
generators: let a derived interface's property override win

joshmouch:
## Summary

An interface that narrows an inherited property was generating the **base** property's type in the rust, go, kotlin and swift generators. Where the base is a discriminated union and the override narrows it to one variant's payload, the generated field is wider than the contract — and the payload struct it should have narrowed to is emitted with `omitDiscriminants`, so a value round-tripped through the wide field loses its tag.

This takes the most-derived declaration of each property.

## Why now, and where the precedent comes from

I wrote the C# version of this in #206, which @connor4312 approved and merged on 2026-08-23. `scripts/generate-csharp.ts` already resolves each property to its most-derived declaration and names the motivating case in a comment (`ToolCallAuthRequiredState.contributor`). So this finishes what that started rather than proposing a new idea.

Two of six generators were correct; four were not. The visible symptom is a disagreement between two first-party clients — .NET emits

```csharp
public required ToolCallMcpContributor Contributor { get; init; }
```

while Go emits

```go
Contributor *ToolCallContributor
```

still carrying the base type's doc comment.

## The Rust half

Rust needs the narrowed payload re-tagged on the wire, since serde's tag lives on the enum and not on the payload struct. A generated serde-`with` module injects the discriminant on serialize.

It also **rejects a foreign discriminant on deserialize**. That part matters: decoding one anyway would re-emit another peer's tag as our own, which is a worse defect than the tag loss being fixed. This mirrors `deserialize_running_tool_call`, which already errors on a mismatched `status`. `narrowed_contributor_rejects_a_foreign_tag` fails without it.

The union registry backing the narrowing lookup is populated by `generateDiscriminatedUnion` itself, so a union cannot be emitted without becoming findable. Asserting membership instead would not hold — `StateAction` is built from runtime-computed variants and can never appear in a static const list.

## Things you'll notice in the diff

- **38 lines of `Uri` → `String`** (19 in Go `commands.generated.go`, 19 in Rust `commands.rs`). This is the loudest thing in the diff and it is not the point of the PR. Cause: `channel` narrows to the literal `'ahp-root://'`. Non-breaking — both are aliases (`type URI = string`, `pub type Uri = String`).
- **A doc-comment cascade** across four files: a narrowed property now carries its *own* doc instead of the base's. That is a second behaviour change and an improvement over the C# precedent, which ships `InitializeParams.Channel` with no summary at all.
- **Hand-written reducer changes** in all four clients. Narrowing the field pushes the wrap/unwrap to the call sites; that is where the per-client diffs come from.

**Why not just drop `omitDiscriminants` in Rust, the way Go does it?** Serde's `tag = "kind"` inserts the key itself, so the payload struct would emit it twice. That is what the `with` module exists to avoid.

## Tests

- rust: 70 across 6 suites, plus `cargo fmt --check` and `cargo clippy -D warnings` clean
- go: all packages ok, `gofmt` clean
- swift: 121
- kotlin: 366
- all four generators regenerate clean

Happy to add a shared round-trip fixture under `types/test-cases/round-trips/` covering a narrowed union payload if you'd prefer cross-client proof over the Rust-only test — that corpus is the natural home for it and I'd rather add it here than leave the other three languages resting on the existing reducer fixtures.

### DIFF
--- a/clients/go/ahp/reducers.go
+++ b/clients/go/ahp/reducers.go
@@ -99,7 +99,10 @@ func toolCallMeta(tc ahptypes.ToolCallState) toolCallCommon {
 	case *ahptypes.ToolCallRunningState:
 		return toolCallCommon{v.ToolCallId, v.ToolName, v.DisplayName, v.Intention, v.ToolInput, v.Contributor, v.Meta}
 	case *ahptypes.ToolCallAuthRequiredState:
-		return toolCallCommon{v.ToolCallId, v.ToolName, v.DisplayName, v.Intention, v.ToolInput, v.Contributor, v.Meta}
+		// ToolCallAuthRequiredState narrows Contributor to the MCP variant; widen it
+		// back to the union toolCallCommon shares with every variant.
+		mcp := v.Contributor
+		return toolCallCommon{v.ToolCallId, v.ToolName, v.DisplayName, v.Intention, v.ToolInput, &ahptypes.ToolCallContributor{Value: &mcp}, v.Meta}
 	case *ahptypes.ToolCallPendingResultConfirmationState:
 		return toolCallCommon{v.ToolCallId, v.ToolName, v.DisplayName, v.Intention, v.ToolInput, v.Contributor, v.Meta}
 	case *ahptypes.ToolCallCompletedState:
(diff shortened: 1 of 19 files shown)


connor4312:
> 38 lines of Uri → String (19 in Go commands.generated.go, 19 in Rust commands.rs). This is the loudest thing in the diff and it is not the point of the PR. Cause: channel narrows to the literal 'ahp-root://'. Non-breaking — both are aliases (type URI = string, pub type Uri = String).

This is non-breaking but it loses semantic meaning. I would much prefer we keep `URI` here

PR #418, opened 2026-08-25
Add optional _meta metadata to SessionActiveClient

toliaqat:
Adds an additive, backward-compatible `_meta` side-channel to `SessionActiveClient`, following the same MCP `_meta` convention already used by `SessionState`, `ToolDefinition`, and `CustomizationBase`.

Hosts can stamp explicit per-connection facts (participant role, read-only status, ...) instead of leaving clients to infer them from an empty `tools` / `customizations` list. The protocol stays neutral: it defines only the carrier plus recommended namespaced-key semantics.

## The field

```ts
export interface SessionActiveClient {
  clientId: string;
  displayName?: string;
  tools: ToolDefinition[];
  customizations?: ClientPluginCustomization[];
  /** Additional provider-specific metadata about this active client ... */
  _meta?: Record<string, unknown>;
}
```

Wire key is `_meta` in every client. All generated artifacts come from `npm run generate` — no hand-edited generated files:

| Client | Generated member |
| --- | --- |
| Rust | `#[serde(rename = "_meta", default, skip_serializing_if = "Option::is_none")] pub meta: Option<JsonObject>` |
| Go | `Meta map[string]json.RawMessage` with `json:"_meta,omitempty"` |
| Kotlin | `@SerialName("_meta") val meta: Map<String, JsonElement>? = null` |
| Swift | `public var meta: [String: AnyCodable]?` + `case meta = "_meta"`; new `init` parameter is defaulted, so it stays source-compatible |
| .NET | `SessionActiveClient.Meta` (nullable, `_meta` JSON name) |
| Schema | `$defs.SessionActiveClient.properties._meta` = `{ "type": "object", "additionalProperties": {} }`, **not** in `required` |

## Backward compatibility

Purely additive and optional. No protocol version bump and no `types/version/registry.ts` change (that registry is action-keyed; `session/activeClientSet` is unchanged at `0.5.0`). Payloads that omit `_meta` decode fine and re-encode with the key omitted rather than emitted as `null`.

## Semantics

Documented in `docs/guide/customizations.md` under a new "Active-Client Metadata" section:

- **Namespace your keys** (reverse-DNS top-level key, payload nested under it) so independent producers cannot collide.
- **Ignore what you do not know**, and keep working when `_meta` is absent.
- **Stamp, do not infer** — an empty `tools` / `customizations` list means only that the client publishes nothing, never that it is read-only.
- **Carry it forward** — `session/activeClientSet` replaces the whole entry, so a dispatcher re-publishing an entry it did not author must preserve the existing `_meta` unless it is deliberately restamping it.

`SessionActiveClient._meta` describes one connection; session-wide facts still belong in `SessionState._meta`.

## Tests

- `types/test-cases/round-trips/044-session-active-client-meta.json` — a `session/activeClientSet` `StateAction` carrying nested `_meta` round-trips verbatim under the `_meta` key.
- `types/test-cases/round-trips/045-session-active-client-meta-absent.json` — a payload predating `_meta` still decodes and re-encodes with the key omitted.
  Both reuse the existing `StateAction` decode entry, so all six client harnesses run them with no harness edits.
- `types/test-cases/reducers/272-…-carries-active-client-meta.json` — upsert insert + replace branches carry and restamp `_meta`.
- `types/test-cases/reducers/273-…-replacement-drops-active-client-meta.json` — whole-entry replacement drops a previously stamped `_meta` (replacement, not a merge).
- `scripts/generate-json-schema.test.ts` — `_meta` stays optional, accepts a namespaced metadata object, rejects a scalar, and a payload without `_meta` still validates.

## Verification

- Root `npm run test`: 438 passed (typecheck, lint, release metadata, changelog, change fragments, `verify:generated`, 100% branch coverage on `types/reducers.ts`).
- `npm run generate`: clean, no drift.
- `cargo test -p ahp-types`: pass, including `roundtrip_corpus`.
- `go test ./...` in `clients/go`: pass.
- TypeScript client `npm test`: 66 passed, including the round-trip corpus.
- `dotnet test AgentHostProtocol.Tests`: 529 passed.
- Kotlin and Swift toolchains were unavailable in this environment. Their generated diff is the same shape as the existing `_meta` fields those suites already cover, and the Swift change only adds a defaulted `init` parameter.

Changelog fragment added at `docs/.changes/20260824-session-active-client-meta.json` (`"type": "added"`, no `targets`, so spec + all clients). No `CHANGELOG.md` files were touched.


types/channels-session/state.ts
```diff
@@ -274,6 +274,28 @@ export interface SessionActiveClient {
...
+   *
+   * A host that assigns participant semantics SHOULD stamp them here
+   * explicitly rather than leaving consumers to infer them from other fields:
+   * an empty {@link SessionActiveClient.tools | `tools`} or
+   * {@link SessionActiveClient.customizations | `customizations`} list means
+   * only that the client publishes nothing, never that it is read-only.
+   * Because `session/activeClientSet` upserts the whole entry, a host that
+   * re-publishes an entry it did not author MUST carry the existing `_meta`
+   * forward unless it is deliberately restamping it.
+   */
```
connor4312, line 288: This comment is kind of long winded, you don't need to rexplain how _meta works, it's already an existing known convention in the protocol

toliaqat: Good call, trimmed it. Dropped the generic namespacing/ignore-unknown/carry-forward boilerplate since that convention is already documented on every other `_meta` field and in the customizations guide, and kept just the stamp-don't-infer point (the actual reason this field exists) with a pointer to the guide for the full semantics.

toliaqat:
Heads up: the `rust` check on this run is red for an infrastructure reason, not a code reason. The job died during **Set up job**, before checkout or any build step ran:

```
Failed to download action 'https://internal-api.service.iad.github.net/repos/dtolnay/rust-toolchain/tarball/4360b5...'
Error: Name or service not known (internal-api.service.iad.github.net:443)
##[error]Failed to download archive ... after 3 attempts.
```

The other 7 checks pass, and this same commit range passed `rust` on the previous push. The only change since then is doc-comment wording.

I re-ran the job's steps locally against `3dac0332` to be sure:

- `npm run generate:rust` then `git diff --quiet -- clients/rust` -> no diff, generated sources up to date
- `cargo fmt --all -- --check` -> clean
- `cargo clippy -p ahp-types -- -D warnings` -> clean
- `cargo test -p ahp-types` -> 8 passed, 0 failed (includes `roundtrip_corpus`)

I do not have permission to re-run jobs on this fork PR (`gh run rerun` returns "Must have admin rights to Repository"). Could someone with write access kick it? Happy to push an empty commit to retrigger instead if that is easier.

PR #433, opened 2026-09-02
chat: make the turnStarted reducer case idempotent

RyanEwen:
Makes the `chat/turnStarted` reducer case a no-op when the turn it names is already active or already in the turn list.

## Why

A client that applies its own dispatched actions optimistically keeps them pending until the host echoes them, and replays every still-pending action over confirmed state on each confirmed action. Some paths never echo, so a pending `chat/turnStarted` can be replayed indefinitely. Today that case unconditionally rebuilds `activeTurn` with empty `responseParts`, so one stranded start wipes the streamed content on every delta. That is the mechanism behind microsoft/vscode#332087, where responses render only at turn end.

The guard also covers a start replayed after the turn completed and was confirmed through a snapshot, which would otherwise resurrect a finished turn as an empty active one. Ignoring a start for a completed turn is the right semantics: restarting a finished turn is `chat/turnResume`'s job, and turn ids are unique.

## What changed

- `types/channels-chat/reducer.ts`: return the state unchanged when `activeTurn.id` or any entry in `turns` already carries `action.turnId`.
- Fixtures `272-chat-turnstarted-for-active-turn-is-no-op` and `273-chat-turnstarted-for-completed-turn-is-no-op`, with `expected` equal to `initial`.
- A change fragment and one paragraph under "Active Turn" in the state-model guide stating the rule.

`tsc` clean, the reducer test suite passes (280), eslint clean, the fragment check passes.

## Not done here

I did not change the Go, Kotlin, Swift and .NET reducers, so the two new fixtures will fail in their fixture-driven suites until they get the same two-line guard. I am happy to add those in this PR if parity in one change is preferred.

*AI disclosure: this comment and the related code were written with the assistance of AI.*

### Public patches and patcher scripts

[Source patch and installed-bundle workaround mapping](https://github.com/RyanEwen/vscode-patches/blob/main/docs/patches/agent-host-protocol-433.md). The [public collection](https://github.com/RyanEwen/vscode-patches) includes the maintained patchers, rollback instructions, regression scripts, and historical snapshots. Build restrictions and exact installer coverage are documented there.

### DIFF
--- a/clients/dotnet/src/AgentHostProtocol/Reducers.cs
+++ b/clients/dotnet/src/AgentHostProtocol/Reducers.cs
@@ -1093,6 +1093,13 @@ public static class Reducers
 
     private static ReduceOutcome ApplyChatTurnStarted(ChatState state, ChatTurnStartedAction a)
     {
+        // A replayed start must not reset the turn in progress or resurrect a finished one.
+        if ((state.ActiveTurn is not null && state.ActiveTurn.Id == a.TurnId)
+            || state.Turns.Exists(t => t.Id == a.TurnId))
+        {
+            return ReduceOutcome.NoOp;
+        }
+
         state.ActiveTurn = new ActiveTurn
         {
             Id = a.TurnId,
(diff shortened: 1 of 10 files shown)


connor4312:
@copilot update reducers in other languages appropriately

connor4312:
Looks like a rust CI failure

PR #441, opened 2026-09-08
docs: propose host idle lifecycle

JasonEtco:
An idle session does not prove that the whole host is safe to stop. This proposal describes optional host-wide idle detection and a replayable stop-ready state for an external controller.

## Approach

The proposal adds a typed initialize capability, optional root lifecycle state, and a server-only replacement action. It describes a host-owned grace period that accounts for all sessions, background work, client tools, and automatic triggers. It also defines admission, replay, restart identity, and conditional compute-stop requirements.

The document follows the existing proposal overviews, with a mental model, role diagram, worked example, and protocol details.

## Review focus

**This is a proposal, not an implemented or agreed contract.** Automatic, irreversible drain is the primary candidate because it closes admission before publishing a stop-ready decision. The tradeoff is that the host can remain unavailable if its controller fails to stop it. An authorized conditional drain request is the alternative.

The drain model, future-trigger policy, authorization boundary, compatibility, and rejection behavior remain open decisions.

## Scope

Documentation only. No changes to protocol types, generated artifacts, executable code, normative specification pages, changelogs, or versions.

<sub>:robot: Written by Copilot</sub>


connor4312:
I'm unclear if this proposal belongs in the protocol. The protocol does not defined how to start an agent host, nor does it have a representation for how to stop (or restart) an agent host -- so signaling when an agent host is safe to stop seems out of scope. For a naive client, they must retain their connection to relevant agent hosts that may be multiplayer because they don't know if another client will connect and do stuff there. For a client who has its own agent host, it already controls the lifecycle and activity.

Can you describe more the scenario you're trying to target?

engstromloboz-sys:
Den mån 21 sep. 2026 23:47Connor Peet ***@***.***> skrev:

PR #449, opened 2026-09-13
docs: list pi-ahp as an AHP server implementation

Qusic:
I started playing with AHP around 0.3 or 0.4. I have also tried agent apps like Paseo and Orca, but I often felt that they had to rebuild a lot of things VS Code already does really well, like diff views, file searching, terminals, and flexible pane and workspace layouts. That is why AHP has always felt promising to me: it lets a harness plug into an existing, mature client instead.

I use pi as my day-to-day harness, so I started building [`pi-ahp`](https://github.com/Qusic/pi-ahp) to connect the two and hopefully make that setup my daily driver. Now that AHP is at 0.9.0 and feels fairly close to stable, I thought this was a good time to add `pi-ahp` to the server implementation lists in the README and docs.

`pi-ahp` is MIT-licensed and available on npm.


connor4312 (approved):
nice!

dmitrivMS (approved):

rzhao271 (approved):

PR #451, opened 2026-09-16
Add repository lists and host preparation discovery

osortega:
## Summary

Represent repository intent as a future-ready list, separate from generic configuration and resolved directories:

```json
{
  "channel": "ahp-session:/new",
  "repositories": [
    { "source": "https://example.org/team/project.git", "revision": "main" }
  ],
  "config": { "mode": "interactive" }
}
```

- Add `RepositorySource { source: URI; revision?: string }` and optional `repositories?: RepositorySource[]` to creation, both configuration queries, and immutable `SessionMetadata` (full state and summaries). A present list is nonempty; omission preserves directory/default creation.
- Discover support on **`InitializeResult.repositoryPreparation`**, not `AgentCapabilities`. Absence means unsupported; `{}` supports one default-revision repository; `revision: true` allows explicit revisions. Without `multipleRepositories: true`, lists longer than one are rejected before preparation.
- Retain the exact requested list, including order and omitted revisions, through Creating/Ready/Failed, summaries, and lost-reply recovery. A source URI is not checkout identity; repeated sources at different revisions remain possible. There is no positional mapping to `workingDirectories`.
- Creation repository lists and nonempty working directories are mutually exclusive. Configuration queries may carry directory context but do not prepare repositories. Resulting directories must fit the selected provider's existing directory capabilities.
- Restore generic configuration descriptions, `disposeSession`, and `root/progress` documentation to main. Keep only scoped preparation ownership guidance and one concise no-clone rule explaining why draft edits should not create checkouts.
- Regenerate schemas and all six clients, including serializer/type registrations and shared conformance fixtures. No aliases, config carrier, repository IDs/catalogue, new RPC, per-repository retry, multi-checkout implementation, dependency change, or version bump.

## Exact revision and base

- **Published protocol pin:** `043bc34700935951efc4528c59e9d3158f463bfc`.
- **Latest fetched main, already included:** `27207eae82cd8b418aef03ff384414d432be88c5`. No additional merge was necessary; existing history was preserved with a normal signed commit and push, without rebase/amend/force.
- Related public client PR: [microsoft/vscode#336361](https://github.com/microsoft/vscode/pull/336361).

## Validation

| Check | Result |
| --- | --- |
| Required `npm run generate`, repeated after staging | Passed; deterministic output |
| Required `npm test` | **467 passed**, including typecheck/lint/metadata/schema checks and **100% reducer coverage** |
| Full Rust `cargo test --workspace` plus workspace fmt/clippy | **88 passed**, including all crates and doctests |
| TypeScript source typecheck and request/wire tests | **38 passed** |
| Go shared wire/reducer corpora and `go vet ./...` | Passed |
| Swift shared wire/reducer corpora | **6 passed** |
| Kotlin shared wire/reducer corpora | **345 passed** |
| .NET library tests | **553 passed** |

The shared corpora contain **59 wire fixtures** and **274 reducer fixtures**. The missing Go and JDK toolchains were restored into a checkout-local task directory and removed after validation. These are protocol/client conformance checks; no live-product validation is claimed.

## CI and reading order

Exact-head CI: [run 35564386401](https://github.com/microsoft/agent-host-protocol/actions/runs/35564386401), **passed — all seven language/check jobs succeeded** on `043bc34700935951efc4528c59e9d3158f463bfc`. [Current commit checks](https://github.com/microsoft/agent-host-protocol/commit/043bc34700935951efc4528c59e9d3158f463bfc/checks). Earlier green runs do not validate this revision.

1. [Host discovery](https://github.com/microsoft/agent-host-protocol/blob/043bc34700935951efc4528c59e9d3158f463bfc/types/common/commands.ts), [creation](https://github.com/microsoft/agent-host-protocol/blob/043bc34700935951efc4528c59e9d3158f463bfc/types/channels-session/commands.ts), [queries](https://github.com/microsoft/agent-host-protocol/blob/043bc34700935951efc4528c59e9d3158f463bfc/types/channels-root/commands.ts), and [metadata](https://github.com/microsoft/agent-host-protocol/blob/043bc34700935951efc4528c59e9d3158f463bfc/types/channels-session/state.ts).
2. [Repository-backed creation contract](https://github.com/microsoft/agent-host-protocol/blob/043bc34700935951efc4528c59e9d3158f463bfc/docs/specification/session-channel.md#repository-backed-creation).
3. [Schema tests](https://github.com/microsoft/agent-host-protocol/blob/043bc34700935951efc4528c59e9d3158f463bfc/scripts/generate-json-schema.test.ts), [request/recovery tests](https://github.com/microsoft/agent-host-protocol/blob/043bc34700935951efc4528c59e9d3158f463bfc/clients/typescript/test/client.test.ts), and [shared fixtures](https://github.com/microsoft/agent-host-protocol/tree/043bc34700935951efc4528c59e9d3158f463bfc/types/test-cases).


engstromloboz-sys (commented):
🫣

types/channels-session/state.ts
```diff
@@ -141,6 +141,10 @@ export interface SessionMetadata {
    * chat that sets none operates against this full set.
    */
   workingDirectories?: URI[];
+  /** Immutable requested source, separate from the host-resolved working directories. */
+  repositorySource?: URI;
```
roblourens: I think these two props should be in an object, and it doesn't make sense to have a revision without a source right?

types/channels-session/state.ts
```diff
@@ -189,7 +193,7 @@ export interface SessionState extends SessionMetadata {
    * this over the session's lifetime.
    */
   defaultChat?: URI;
-  /** Session configuration schema and current values */
+  /** Provider-specific session configuration schema and current values. */
```
roblourens: Why this change?

types/channels-session/commands.ts
```diff
@@ -104,6 +114,9 @@ export interface CreateSessionParams extends BaseParams {
  * Disposes a session and cleans up server-side resources.
  *
  * The server broadcasts a `root/sessionRemoved` notification to all clients.
+ * Disposal MUST NOT erase a shared checkout or uncommitted user changes.
+ * Repository cleanup remains host-owned; ending a client's wait or subscription
```
roblourens: Why this change? Not sure I agree or understand it fully

types/channels-session/state.ts
roblourens: Would we support multiple repo checkouts?

types/channels-root/state.ts
```diff
@@ -106,6 +106,8 @@ export interface AgentInfo {
  * @category Root State
  */
 export interface AgentCapabilities {
+  /** The host accepts typed repository inputs for session creation and configuration queries. */
+  repositorySource?: RepositorySourceCapability;
```
roblourens: We discussed agent-specific capability vs host-specific, why did you put it on the agent?

types/channels-root/notifications.ts
```diff
@@ -175,6 +175,8 @@ export interface SessionSummaryChangedParams {
  * - Like all notifications this is ephemeral and is **not** replayed on
  *   reconnect. A client that never receives the terminal frame SHOULD expire
  *   the indicator after an idle timeout.
+ * - Completion of reported work does not establish session readiness.
```
roblourens: Not totally understanding why this comment is on root/progress

types/channels-root/commands.ts
```diff
@@ -79,6 +79,9 @@ export interface ListSessionsResult extends PaginatedResult {
  * the full current property set (not a delta). The returned `values` contain
  * server-resolved defaults to pass to `createSession`.
  *
+ * This command MUST NOT clone or prepare a repository. Repository context
```
roblourens: Not sure this is necessary, do you see this being a likely issue?

types/channels-session/state.ts
osortega: Addressed in `043bc34700935951efc4528c59e9d3158f463bfc`.

Grouped the inputs as `repositories: RepositorySource[]` on creation, both config queries, and shared session metadata. Every entry requires `source`; `revision` cannot stand alone. A present list is nonempty. `InitializeResult.repositoryPreparation.multipleRepositories` gates lengths greater than one; absent/false requires rejection before preparation. The complete list survives Creating/Ready/Failed and summaries. Sources are not checkout identities or deduplication keys, and there is no positional directory mapping or multi-checkout implementation.

types/channels-session/state.ts
osortega: Addressed in `043bc34700935951efc4528c59e9d3158f463bfc`.

Restored the generic configuration descriptions to main, including `SessionState.config`, creation config, and resolution config. Repository selection now lives in typed `repositories` beside `config`; it does not narrow generic configuration or add repository metadata to `SessionConfigSchema`.

types/channels-session/commands.ts
osortega: Addressed in `043bc34700935951efc4528c59e9d3158f463bfc`.

Removed the repository-specific disposal paragraph and the broad cancellation/disposal guarantees. The generic `disposeSession` documentation now matches main. Only scoped ownership guidance remains in repository-backed creation: the host owns resources it creates, and a source URI does not establish ownership of an existing/shared checkout. No new disposal rules or runtime behavior changes are introduced.

types/channels-root/state.ts
osortega: Addressed in `043bc34700935951efc4528c59e9d3158f463bfc`.

Moved discovery to `InitializeResult.repositoryPreparation: RepositoryPreparationCapabilities` and removed the agent capability entirely. An absent object is unsupported; `{}` supports one default-revision repository; `revision: true` opts into explicit revisions and `multipleRepositories: true` into longer lists. Preparation belongs to the host; its resulting directories must still fit the selected provider's existing directory capabilities. `AgentCapabilities` now matches main.

types/channels-root/notifications.ts
osortega: Addressed in `043bc34700935951efc4528c59e9d3158f463bfc`.

Removed the added readiness bullet and root-channel paragraph, so generic `root/progress` comments and documentation match main again. Preparation progress/readiness guidance is scoped to repository-backed creation instead; session state remains the durable record of the repository list, directories, and outcome.

types/channels-root/commands.ts
osortega: Addressed in `043bc34700935951efc4528c59e9d3158f463bfc`.

Kept one concise rule covering both configuration-query commands, with the rationale: editing a draft should not create checkouts. Configuration resolution/completions can be called repeatedly before the user chooses Create, so repository lists and existing directory context are only query context. Preparation happens at creation. Removed the repeated per-field no-clone/agent-capability wording.

types/channels-session/commands.ts
```diff
@@ -26,6 +27,9 @@ import type {
  * updates. The server also broadcasts a `root/sessionAdded` notification to all
  * clients.
  *
+ * Repository preparation MUST finish before `session/ready` or executing turns.
+ * Clients recover the outcome from session state, not progress notifications.
```
roblourens: > "recover the outcome from session state, not progress notifications"

This is still confusing, what is this meant to communicate?

types/channels-root/commands.ts
```diff
@@ -130,6 +133,13 @@ export interface ResolveSessionConfigParams extends BaseParams {
   provider?: string;
   /** Working directory for the session */
   workingDirectory?: URI;
+  /**
+   * Non-empty repository context, subject to
+   * {@link InitializeResult.repositoryPreparation}. May accompany `workingDirectory`.
+   *
+   * @minItems 1
```
roblourens, line 137: This `@minItems` seems unnecessary when the property is optional

types/common/commands.ts
```diff
@@ -295,6 +301,24 @@ export interface InitializeResult {
...
 }
 
+/**
+ * Repository preparation supported by this host, independent of the selected
+ * agent. Resulting working directories must still fit that agent's existing
+ * directory capabilities.
+ *
+ * @category Commands
+ */
+export interface RepositoryPreparationCapabilities {
```
roblourens, line 310: Should this advertise support for `git` specifically vs other types of repos? And do you feel that the sub-capability for `revision` is necessary?

types/channels-root/commands.ts
roblourens: This comment is still here- I'm not seeing why it's necessary

connor4312:
Some questions/comments:

- It'd be nice if we can extend this later to have first class worktree support for local AH! Also thinking about about worktree-style operations that aren't necessarily using git worktrees, like [rift](https://github.com/anomalyco/rift) and similar
- We say that repositories are present "instead of an explicit `workingDirectories` list", but this seems to prevent the ability to create sessions in subdirectories of repositories, which is a popular monorepo scenario
- We have existing capabilities around the mutability of working directories--adding or changing them. Are these operations similarly 'banned' for any session with a repositories list?
- We should document (in .md) the expectations around what hosts do here when presented with repositories, there appears to be some subtlties in the lifecycle about when the repository is "prepared"

types/channels-session/state.ts
```diff
@@ -141,6 +153,12 @@ export interface SessionMetadata {
    * chat that sets none operates against this full set.
    */
   workingDirectories?: URI[];
+  /**
+   * Immutable repository inputs accepted at creation.
+   *
+   * @minItems 1
+   */
+  repositories?: RepositorySource[];
```
roblourens: When I create a session for a repo, do I then get a working directory for the clone? Or in a local scenario, does the host not tell the client what the directory is? And could that cause confusion either way with things like wanting to know which repo is associated with which working directory?

types/channels-root/commands.ts
osortega: Removed the repeated no-checkout wording from the declarations. Configuration-query versus creation behavior is now described in the Markdown spec.

types/channels-root/commands.ts
osortega: Optional allows omission for directory/default creation. @minItems 1 rejects a supplied empty list, so [] cannot silently request a default instead of preparing a source.

types/channels-session/commands.ts
osortega: Removed that vague wording. The spec now says to wait for session/ready, not progress or command success. A duplicate create remains SessionAlreadyExists: it is not proof of successful creation and must not trigger automatic pending turns. Ordinary reconnect is unchanged.

types/common/commands.ts
osortega: Kept the existing host-level repositoryPreparation shape and revision/multipleRepositories flags in this update. No Git-specific capability split or new preparation backend is introduced here.

types/channels-session/state.ts
osortega: Yes: workingDirectories carries the actual selected directory. Opted-in clients also receive that entry’s repo source and origin (local/repo/worktree, with mainWorktree for worktrees). Association is on the URI-keyed record, not array order. URI-only clients still receive the resolved paths. Implemented in 0ceeb99b.

osortega:
@connor4312 Updated in 0ceeb99b: repositories[].subdirectory selects a monorepo folder; resolved directories carry host-owned source/provenance metadata. Existing directory mutation and primary-slot rules are unchanged. The [spec](https://github.com/microsoft/agent-host-protocol/blob/0ceeb99be0ac7821d1f4ca887ed5771a6e15391a/docs/specification/session-channel.md#repository-backed-creation) covers preparation, readiness, compatibility, and lost creation responses. This describes results; it does not add generic worktree creation or a Rift backend.

PR #453, opened 2026-09-16
chat: Add source attribution to agent responses

### LINKED ISSUE #452: Support source attribution in agent responses
## Problem

An agent can explain an answer using web pages, files, or retrieved documents, but AHP has no shared way to say which sources support which parts of that answer.

Ordinary Markdown links help people open sources, but they do not give clients enough information to show a source preview or highlight the text it supports. Information attached only to a live action is also missing when another client joins or loads the conversation history.

## Proposal

Add an optional attribution response part to the chat channel. It would identify an earlier text part, list its sources, and link text ranges to those sources.

This should work with any agent or harness. It should not require a particular model, search tool, website, local filesystem, or user interface.

## Expected behavior

- The host can stream the answer first and add attribution before the turn ends.
- Sources can describe web pages, files, and documents, with optional titles, excerpts, and locations.
- Text ranges have a precise, language-independent meaning, including for non-ASCII text.
- Attribution survives turn completion, snapshots, reconnection, and history loading.
- A client that does not show attribution can still display the original answer normally.
- Attribution does not grant permission to fetch a resource or claim that the answer has been independently verified.

## Scope

Add the protocol types, explain the behavior with a small example, regenerate the client libraries and schemas, and cover the wire format and shared-state behavior with the existing cross-language tests.

Search planning, search progress, ranking, and the appearance of citation badges or source panels are outside this change.

lramos15:
## Problem

An agent may base an answer on web pages, files, or documents. AHP currently has no shared way to connect those sources to the parts of the answer they support. That information should remain available when another client joins, reconnects, or loads older turns.

Closes #452.

## What changed

- Added an optional `attribution` response part. It points to an earlier Markdown or reasoning part and lists supporting sources and text ranges.
- Kept the existing streaming flow: send the answer first, then append attribution with `chat/responsePart` before ending the turn. No new command or action is needed.
- Added optional source titles, URIs, short quotations, and text or page locations. Sources do not have to be public URLs or local files.
- Defined how identifiers, UTF-16 positions, overlapping ranges, turn resume, and unknown location kinds behave.
- Regenerated the schemas and client types, updated client part-ID handling, and added shared round-trip and reducer tests.
- Added a plain-English guide and a changelog fragment.

## Compatibility and scope

Clients can ignore attribution and continue showing the original answer. The data stays in chat state for snapshots and history. It does not grant access to a source or claim that the answer has been independently verified.

The change is not tied to a particular agent, model, harness, or search tool. Search planning, live search progress, ranking, and citation/source-panel UI are not included. No protocol or package version was changed.

## Validation

All passed:
- `npm run generate`
- `npm test` (459 tests; 100% reducer branch coverage)
- `npm run docs:build`
- TypeScript client tests and build
- Rust formatting, Clippy, workspace tests, and typed attribution round trips
- Swift package tests
- Go client tests and `go vet`
- Kotlin client tests using an isolated JDK 17 container
- .NET solution formatting, Release build, and all 543 tests using an isolated .NET 10 SDK, .NET 8 runtime, and Node 24 container

The shared fixtures cover attribution after streaming, Unicode and multi-line ranges, overlapping spans, whole-answer attribution, history loading, cancellation, error/resume, snapshots, and unknown source-location kinds.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


Giuspepe (approved):

dmitrivMS (approved):

types/channels-chat/state.ts
```diff
@@ -931,6 +932,133 @@ export interface ReasoningResponsePart {
...
+ * text delta and before the turn ends. At most one attribution part may target
+ * a given part. Neither the target text nor its attribution changes afterward,
+ * including when the turn resumes; further output uses new part identifiers.
+ *
+ * Clients MAY show inline citations or a source list. Clients that do not
+ * support attribution can ignore this part and still render the original text.
+ *
+ * @category Response Parts
+ */
+export interface AttributionResponsePart {
```
connor4312, line 948: Do you feel a tendency towards this approach versus interpolating the attribution between or within markdown parts? Do we actually get these late from model providers?

lramos15: We do get them later from model providers, but I discussed with Rob and we're going to hold off on this approach for now because I was aiming to improve web search with this but turns out web search was just broken in the SDK which is originally why it was bad

